PR #2, opened 2024-12-29
Allow users to invoke apps by name

bradhe:
The Tower CLI previously allowed you to supply an optional directory as an argument such that you could be build/deploy/run a Tower app from anywhere in your filesystem. We've pushed that into a named argument (`--dir`) such that you can invoke `tower run <app name>`.


sankroh (approved):
👍

PR #4, opened 2025-01-23
feat: Add in upgrade nudge if newer version is available.

sankroh:


crates/tower-cmd/src/lib.rs
```diff
@@ -23,8 +27,41 @@ impl App {
...
+        Ok(None)
+    }
+
     pub async fn run(self) {
+        let mut cmd_clone = self.cmd.clone();
         let matches = self.cmd.get_matches();
+
+        // Check for newer version
+        if let Ok(Some(latest_version)) = Self::check_latest_version().await {
+            let current_version = env!("CARGO_PKG_VERSION");
```
bradhe, line 52: Looks good, since we reach in to `env!` to get the version here and in the `version` cmd, we should centralize this at some point into a `version` crate.

crates/tower-cmd/src/lib.rs
```diff
@@ -23,8 +27,41 @@ impl App {
...
+                latest_version, current_version).yellow());
+                eprintln!("{}", "To upgrade, run: pip install --upgrade tower-cli\n".yellow());
+            }
+        }
+
+        // If no subcommand was provided, show help and exit
+        if matches.subcommand().is_none() {
+            cmd_clone.print_help().unwrap();
+            std::process::exit(0);
+        }
```
bradhe, line 64: Oh man, thanks for fixing this. Should it get pushed down to after log gets initialized? This could hide some environment initialization errors that would pop up, and/or maybe behavior changes of this commend (somehow) after initialization.

On the exit code, by the way, returning something non-zero here would be good. Currently Rust uses exit code 101 when you supply an unknown command. I checked `psql` as another example, it uses exit code 2 for an unknown command.

crates/tower-cmd/src/lib.rs
```diff
@@ -23,8 +27,41 @@ impl App {
...
+            .get("https://pypi.org/pypi/tower-cli/json")
+            .send()
+            .await?;
+            
+        if resp.status().is_success() {
+            let json: Value = resp.json().await?;
+            if let Some(version) = json.get("info").and_then(|info| info.get("version")).and_then(|v| v.as_str()) {
+                return Ok(Some(version.to_string()));
+            }
+        }
```
bradhe, line 42: Debug message if this fails might be good. Another argument for pushing all of this to after initialization (see other comment).

bradhe (changes requested):
Looks good generally! Few minor things to address.

bradhe:
Also can you open this against `develop` pretty please?

crates/tower-cmd/src/lib.rs
sankroh: good idea.

crates/tower-cmd/src/lib.rs
sankroh: 👍

bradhe (approved):
Looks great, good addition to the CLI!

PR #6, opened 2025-01-30
Handle incorrect user input

sankroh:
- Fixes unreachable exception when user enters a bad or invalid command.
- Change to error messages to make it a bit friendlier.


bradhe (approved):
Looks good to me, good catches all around.

PR #7, opened 2025-01-31
Handle API errors correctly

sankroh:
- Some API errors have sub errors in an `items` field, make sure the api client is able to handle that.


crates/tower-cmd/src/output.rs
```diff
@@ -100,6 +100,15 @@ pub fn runtime_error(err: tower_runtime::errors::Error) {
 pub fn tower_error(err: tower_api::TowerError) {
     let line = format!("{} {}\n", "Error:".red(), err.description.friendly);
     io::stdout().write_all(line.as_bytes()).unwrap();
+
+    // Handle any nested validation errors if present
+    if let Some(items) = err.items {
+        writeln!(io::stdout(), "\n{}", "Exception details:".yellow()).unwrap();
```
bradhe, line 106: We call this an "Error" above, perhaps it should be "Error details" here?

bradhe (approved):
One small suggestion. Good otherwise.

PR #8, opened 2025-01-31
Prompt user to create app if app doesn't exist.

sankroh:
- Add schedules from Towerfile to Apps when creating them
- Add an App description field to Towerfile, and syncs it with Apps
- Add a utility function to check if app exists and prompt the user to create it, if it does not.


bradhe (approved):
LGTM, good additions!

PR #9, opened 2025-02-10
Change API client to client based on OpenAPI spec.

sankroh:


crates/tower-cmd/src/secrets.rs
```diff
@@ -67,61 +72,71 @@ pub fn secrets_cmd() -> Command {
...
+                        output::table(headers, data);
+                    },
+                    default_api::ListSecretsSuccess::UnknownValue(_) => {
+                        output::failure("Received unknown response format from server");
+                    }
+                }
+            }
+        },
+        Err(err) => output::tower_error(err),
+    }
```
bradhe, line 132: There's a lot going on here, maybe it should be broken up more? Separate the outer pattern matching from the formatting perhaps?

crates/tower-cmd/src/secrets.rs
```diff
@@ -130,21 +145,57 @@ pub async fn do_create_secret(_config: Config, client: Client, args: &ArgMatches
...
+    }).await {
+        Ok(key_response) => {
+            if let Some(key_success) = key_response.entity {
+                match key_success {
+                    default_api::DescribeSecretsKeySuccess::Status200(key_response) => {
+                        let public_key = rsa::RsaPublicKey::from_pkcs1_pem(&key_response.public_key)
+                            .unwrap_or_else(|_| {
+                                spinner.failure();
+                                output::die("Failed to parse public key");
+                            });
```
bradhe, line 160: Can you push this into the `crypto` crate? It should live with all that.

crates/tower-cmd/src/util.rs
```diff
@@ -1,21 +1,60 @@
...
+                                        content: create_err.to_string(),
+                                        entity: None
+                                    }
+                                ))
+                            }
+                        } else {
+                            Err(err)
+                        }
+                    } else {
+                        Err(err)
```
bradhe, line 57: This too is a bit crazy. I'd see if you can have Claude clean this up a bit to make it more clear what's happening in here.

crates/tower-cmd/src/version.rs
```diff
@@ -1,14 +1,14 @@
...
+use tower_api::apis::configuration::Configuration;
 use config::Config;
 
 pub fn version_cmd() -> Command {
     Command::new("version")
         .about("Print the current version of Tower")
 }
 
-pub async fn do_version(_config: Config, _client: Client) {
+pub async fn do_version(_config: Config, _configuration: &Configuration) {
```
bradhe, line 11: Probably don't need this here.

openapi/generate_tower_client.sh
```diff
@@ -0,0 +1,15 @@
...
+echo "Generating API code..." && \
+podman run --rm \
+-v ${PWD}:/local openapitools/openapi-generator-cli generate \
+-c /local/generator_config.yaml && \
+
+echo "Copying generated files to target directory..." && \
+rsync -av --exclude={'.gitignore','.openapi*','.travis.yml','*.sh'} tower-api/ ../crates/tower-api/ && \
+
+echo "Cleaning up..." && \
+rm -r openapi.json tower-api
```
bradhe, line 15: Should this go in the `scripts` directory?

bradhe (changes requested):
Okay, there's a few things to address that I think are important but the meat of it seems to be there.

1. The `deploy` command doesn't seem to actually deploy anything, and we need a way for it to report out progress for the progress bar to update.
2. Maintaining separate config + API config will make it such that the two objects can get out of sync. Everywhere we need an API config, the data for it comes from `config::Config`. Let's make it such that `config::Config` returns an API configuration!
3. There's a lot of clean up that can be done on importing to remove lots of boilerplate.
4. Some of the pattern matching has gotten a bit hairy with all the control flow tangled up with the logic. I'd say run it through Claude and have it help clean that up a bit, to make the control flow and logic more separated.
5. Minor shuffling of some things (errors in `output` package, deserializing public RSA keys in the `crypto` package) to align with existing semantics.

Like I said, the big chunks are in there's just a few bugs and some cleanup to be done.

crates/config/src/lib.rs
```diff
@@ -19,13 +20,16 @@ pub use towerfile::Towerfile;
 pub struct Config {
     pub debug: bool,
     pub tower_url: Url,
+    #[serde(skip_serializing, skip_deserializing)]
+    pub api_configuration: Option<Configuration>
```
bradhe, line 24: Do we need to persist it? We can just return a config directly.

crates/tower-cmd/src/apps.rs
```diff
@@ -47,170 +63,228 @@ pub fn apps_cmd() -> Command {
...
+                output::failure(&format!("{}: {}", err.status, err.content));
+            } else {
+                output::failure(&format!("Unexpected error: {}", err));
+            }
         }
     }
 }
 
-pub async fn do_create_app(_config: Config, client: Client, args: &ArgMatches) {
+pub async fn do_create_app(_config: Config, configuration: &Configuration, args: &ArgMatches) {
```
bradhe, line 235: Looks like this still needs to be updated.

bradhe (approved):
There's a number of things to address from a cleanup perspective, but let's land this to develop and do the cleanup there! Thanks, this is great so far.

Issue #11, opened 2025-02-15
tower apps create crashes


alitouka:
The `tower apps create` command crashes in Tower CLI 0.1.24 (see attached screenshot) . It seems to be a new issue that did not exist in 0.1.23. The stack trace points to [this line](https://github.com/tower/tower-cli/blame/72d817f66733f42c13657712aed1b88e9de33535/crates/tower-cmd/src/apps.rs#L178) . IIUC that code obtains the value of the `schedule` command line arg. When I tried to add it to the command line (as in `tower apps create --name "hello-world" --description "Test" --schedule 1h`), I got another error: `error: unexpected argument '--schedule' found` . For now, I can get around this issue by downgrading tower-cli to 0.1.23.

<img width="1524" alt="Image" src="https://github.com/user-attachments/assets/19d3bd7e-564d-4479-8e52-60418ccdeb25" />

sankroh:
Thanks for reporting! We’re looking into it.

alitouka:
@bradhe @sankroh thanks for the quick fix! I upgraded to 0.1.25 and confirmed that the issue has been resolved.

PR #12, opened 2025-02-15
Fix for create app issue

sankroh:
Fix for create app issue #11 , add in optional schedule parameter.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -321,7 +321,7 @@ dependencies = [
 
 [[package]]
 name = "config"
-version = "0.1.23"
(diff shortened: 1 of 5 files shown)


bradhe (changes requested):
Actually took an alternate approach to this, in removing the ability to schedule at app creation time all together. It doesn't make sense to specify a schedule for an app outside the deployment flow. We don't allow you to specify this anywhere else. See #13 .

I like the tooling for version management, though. Can you cherry pick that across?

sankroh:
Done.

PR #14, opened 2025-02-17
Add in version helper

sankroh:
- Add in tooling to help bump versions uniformly.
- Backport version bump commit to main

### DIFF
--- a/bump-version.sh
+++ b/bump-version.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+# Function to increment version
+bump_version() {
+    local version=$1
+    local type=$2
+    
+    # Split version into major, minor, patch
+    IFS='.' read -r major minor patch <<< "$version"
+    
+    case $type in
+        major)
+            major=$((major + 1))
+            minor=0
+            patch=0
+            ;;
+        minor)
+            minor=$((minor + 1))
+            patch=0
+            ;;
+        patch)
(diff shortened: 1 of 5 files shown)


bump-version.sh
```diff
@@ -0,0 +1,70 @@
...
+# Update version in pyproject.toml
+if [ "$(uname)" == "Darwin" ]; then
+    # macOS sed
+    sed -i '' "s/^version = \"$current_version\"/version = \"$new_version\"/" pyproject.toml
+else
+    # GNU sed
+    sed -i "s/^version = \"$current_version\"/version = \"$new_version\"/" pyproject.toml
+fi
+
+echo "Version bumped successfully in Cargo.toml and pyproject.toml!"
```
bradhe, line 70: Sorry, I just remembered one last thing: Part of the process also includes propagating the new version to the Cargo.lock file. You can do this by doing a `cargo build` after the version is bumped. This is required, otherwise the subsequent builds fail due to an outdated Cargo.lock file!

Let's add that as the very last step here. Finally, let's move this to a `scripts` directory, to match conventions of other repos.

bradhe (changes requested):
One more tiny change that I had forgotten about! If I had documented this process (my bad), it wouldn't be such an issue...

bradhe (approved):
Great, thanks @sankroh !

PR #15, opened 2025-02-17
chore: Add git things

sankroh:


bradhe (changes requested):
Minor typo. Can you also make sure this only gets executed on main, pretty please?

PR #23, opened 2025-03-30
Migrate from `tower-cli` to `tower` project

bradhe:
This PR migrates all the configuration for this project such that we use `tower` instead of `tower-cli` in PyPI.


sankroh (approved):
LGTM

PR #24, opened 2025-04-03
Package the Towerfile with the bundle sent to Tower

bradhe:
This PR makes it such that we always package the Towerfile that we built the package from along with the rest of the content. This allows us to track changes to the Towerfile between deployments and all that. Obviously, on the services' side, the relevant content will be encrypted at rest.


sankroh (approved):
👍

PR #26, opened 2025-04-07
Collect and forward Setup-related logs as well as program logs

bradhe:
This PR introduces the ability for setup-related logs to be collected and forwarded. It slightly changes the semantics of log collection, too, by starting the log collection process even before the calling programs ask for log output. It buffers all logs in memory 'till they start getting pulled out.

### DIFF
--- a/crates/tower-runtime/src/local.rs
+++ b/crates/tower-runtime/src/local.rs
@@ -16,7 +16,10 @@ use crate::{
 use tokio::{
     sync::oneshot,
     sync::oneshot::error::TryRecvError,
-    sync::mpsc::channel,
+    sync::mpsc::{
+        UnboundedSender,
+        unbounded_channel,
+    },
     process::Command,
 };
 
@@ -25,7 +28,6 @@ use tokio::{
     io::{AsyncRead, BufReader, AsyncBufReadExt},
     time::{timeout, Duration},
     sync::Mutex,
-    sync::mpsc::Sender,
     process::Child, 
 };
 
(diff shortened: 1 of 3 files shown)


crates/tower-runtime/src/local.rs
```diff
@@ -191,13 +207,25 @@ impl App for LocalApp {
...
             if let Ok(mut child) = res {
+                // Let's also send our logs to this output channel.
+                let stdout = child.stdout.take().expect("no stdout");
+                tokio::spawn(drain_output(FD::Stdout, Channel::Setup, output_sender.clone(), BufReader::new(stdout)));
+
+                let stderr = child.stderr.take().expect("no stderr");
+                tokio::spawn(drain_output(FD::Stderr, Channel::Setup, output_sender.clone(), BufReader::new(stderr)));
+
+                log::debug!("waiting for dependency installation to complete");
+
```
sankroh, line 228: Couple of things here to nitpick.
- If the child exits before the drain_output processes we would lose some output? or fail to see errors?
- Maybe some sort of timeout here would be good, in case `pip install` gets in a hung state waiting for something?

sankroh (approved):
👍

crates/tower-runtime/src/local.rs
bradhe: Good idea re: Timeout, I'll add a change for that. re: output, any output should be buffered as I understand so we should get it. In this case, I don't think execution starts 'till we call `wait` on the process.

PR #32, opened 2025-04-16
Add feature flags

sankroh:
- Add in support for optional dependencies
- Ensure that the intended usage pattern is communicated to the user
- Add docs

ref: TOW-286


bradhe (approved):
Good idea, looks good to me.

PR #42, opened 2025-05-05
Add additional methods to the Table function

bradhe:
This PR introduces a collection of new functions to the Table interface to support common functionality.

* `upsert` allows users to upsert records based on the specified `join_cols` (which are columns be names)
* `delete` allows users to delete records. You can supply a set of predicates (filters) that are `pyarrow.compute.Expression` for matching what records to delete
* `rows_affected` will return information about the rows affected in your operations
* `schema` will return the PyArrow schema
* `to_polars` returns a `polars.LazyFrame` that can be used for creating data frames. This is meant to be a replacement for the `read` function that will be deprecated.

## TODO

- [x] Add some simple tests for `Table` and friends


src/tower/_tables.py
```diff
@@ -31,14 +46,104 @@ def read(self) -> pl.DataFrame:
...
+
+        # NOTE: There is, unfortunately, no way to get the number of rows
+        # deleted besides comparing the two snapshots that were created.
+
+        return self
+
+    def schema(self) -> pa.Schema:
+        # We take an Iceberg Schema and we need to convert it into a PyArrow Schema
+        iceberg_schema = self._table.schema()
+        return convert_iceberg_schema(iceberg_schema)
```
datancoffee, line 145: you can get the Arrow schema by calling as_arrow()
See https://py.iceberg.apache.org/reference/pyiceberg/schema/#pyiceberg.schema.Schema.as_arrow
Unless you want to implement some special mapping, of course

You can reduce this to:

iceberg_schema = self._table.schema()
return iceberg_schema.as_arrow()

src/tower/utils/pyarrow.py
```diff
@@ -53,6 +65,52 @@ def arrow_to_iceberg_type(arrow_type):
     else:
         raise ValueError(f"Unsupported Arrow type: {arrow_type}")
 
+
+def iceberg_to_arrow_type(iceberg_type):
```
datancoffee, line 69: you can get the Arrow schema by calling as_arrow()
See https://py.iceberg.apache.org/reference/pyiceberg/schema/#pyiceberg.schema.Schema.as_arrow
Unless you want to implement some special mapping, of course

You can reduce this to:

iceberg_schema.as_arrow()

src/tower/utils/pyarrow.py
```diff
@@ -65,7 +123,55 @@ def convert_pyarrow_field(num, field) -> types.NestedField:
...
+    return pa.field(name, arrow_type, nullable=not field.required)
+
+
 def convert_pyarrow_schema(arrow_schema: pa.Schema) -> IcebergSchema:
     """Convert a PyArrow schema to a PyIceberg schema."""
     fields = [convert_pyarrow_field(i, field) for i, field in enumerate(arrow_schema)]
     return IcebergSchema(*fields)
+
+
+def convert_iceberg_schema(iceberg_schema: IcebergSchema) -> pa.Schema:
```
datancoffee, line 141: we might not need all of this

you can get the Arrow schema by calling as_arrow()
See https://py.iceberg.apache.org/reference/pyiceberg/schema/#pyiceberg.schema.Schema.as_arrow
Unless you want to implement some special mapping, of course

You can reduce this to:

iceberg_schema.as_arrow()

src/tower/_tables.py
```diff
@@ -31,14 +46,104 @@ def read(self) -> pl.DataFrame:
...
         return pl.scan_iceberg(self._table).collect()
 
-    def insert(self, data: pa.Table):
+
+    def to_polars(self) -> pl.LazyFrame:
+        """
+        Converts the table to a Polars LazyFrame. This is useful when you
+        understand Polars and you want to do something more complicated.
+        """
+        return pl.scan_iceberg(self._table).collect()
```
datancoffee, line 55: Shouldn't this be without the .collect() call?

return pl.scan_iceberg(self._table)

We want to return a LazyFrame, ya?

datancoffee (commented):
Looks good! 
Two suggestions: don't call collect() in lazy read, and use as_arrow() to get the arrow schema

src/tower/_tables.py
bradhe: Good catch, fixed. Thanks!

src/tower/_tables.py
bradhe: Great, thanks. Converted this to use PyIceberg's methodology instead of our own.

PR #44, opened 2025-05-08
v0.3.13 Release

bradhe:
- Upgrade Rust client, fixing bugs in how secrets are deleted and how they are exported.
- Upgrade Python client while we're at it
- Bump version to v0.3.13
- Fix login bug when users can't launch a browser


sankroh (approved):
👍

PR #47, opened 2025-05-09
Updated docstrings in _tables

datancoffee:
Updated docstrings only, so that I could auto-generate the Markdown for Docs.


bradhe (approved):
Nice addition!

PR #48, opened 2025-05-10
Add a simple `wait_for_runs` function

bradhe:
This new function allows users to wait for multiple runs that have been started at the same time. Currently, it just serially waits for each run to complete. In the future, we'll want to provide an async interface into this functionality.


src/tower/_client.py
```diff
@@ -119,3 +119,21 @@ def wait_for_run(run: Run) -> None:
...
+
+    Args:
+        runs (List[Run]): A list of runs to wait for.
+
+    Raises:
+        RuntimeError: If there is an error fetching the run status or if any
+            of the runs fail.
+    """
+    for run in runs:
+        wait_for_run(run)
```
datancoffee, line 139: This implementation is fine for the success case (all runs that we are waiting for are successful).

But in control flows that I am familiar with (e.g. Databricks or Google Dataflow), when people are waiting for 3 child apps, if one of them ends with a failure, the entire parent pipeline fails too.

I think this implementation is ok for v1, but we should think about 
- Returning run status in the wait_for_run()
- in the wait_for_runs(), check for run status of the children, and fail the function if ANY of the child runs fails.

datancoffee:
I updated the docstrings

src/tower/_client.py
bradhe: This is a really good point. What do you think the default behavior should be? I was considering adding a flag that controls the failure behavior.

```
def wait_for_runs(runs: List[Runs], raise_on_failure: bool = True) -> None:
```

And if any of these fail, we'd raise an exception (and stop the parent pipeline). Perhaps that should be the default behavior? Likewise, perhaps it should return a tuple of `(successful_runs, failed_runs)` instead of `None`? What do you think?

datancoffee: @bradhe If we look at how other orchestrators do it, they have a pretty complex framework for defining what to do in case of task failures, for a good reason (to allow users to reuse partial calculations and avoid rerunning the entire pipeline). 

Airflow:
https://medium.com/@jaywang.recsys/apache-airflow-recovery-from-a-tasks-e66234b6d671
Dagster:
https://docs.dagster.io/guides/deploy/execution/run-retries

I think it would be too early to introduce these retry mechanisms, but at least we should tell users if the child app was successful or not. Then they can decide in the orchestrator app what to do: retry or abort.

The raise_on_error where we would throw an exception would be helpful in use cases when the user wants to stop executing the parent app without dealing with child app statuses. But if we add this parameter, it needs to be set to False by default, otherwise users won't be able to deal with failures themselves.

So, I think this means, 

Option 1: the user deals with retries themselves and also decides whether to abort the parent app themselves.
def wait_for_run(run: Run) -> Bool:
   returns success or failure

def wait_for_runs(runs: List[Run]) -> (successful_runs, failed_runs):
   returns a list of successes/failures


Option 2: the user still has to deal with retries themself, but can tell us whether to abort the parent app or not.
def wait_for_run(run: Run, raise_on_failure: bool = False) -> Bool:
   returns success or failure

def wait_for_runs(runs: List[Runs], raise_on_failure: bool = False) -> (successful_runs, failed_runs):
   returns a list of successes/failures, or raises an exception on first failure among children, if user decided to do so

datancoffee (commented):
@bradhe small nit.
With the new timeout related to the overall longevity of any child run being 1 day, perhaps WAIT_TIMEOUT should be renamed to something like POLL_FREQUENCY 

# WAIT_TIMEOUT is the amount of time to wait between requests when polling the
# Tower API.
WAIT_TIMEOUT = 2

src/tower/_client.py
```diff
@@ -171,4 +254,8 @@ def wait_for_runs(runs: List[Run]) -> None:
             of the runs fail.
     """
     for run in runs:
-        wait_for_run(run)
+        wait_for_run(
```
datancoffee, line 257: @bradhe Wait, aren't we supposed to accumulate a list of successful and failed runs? 
and return tuple[List[Run], List[Run]]

src/tower/_client.py
```diff
@@ -134,26 +189,51 @@ def wait_for_run(run: Run) -> None:
...
+                    return True
+                elif _is_failed_run(desc):
+                    if raise_on_failure:
+                        raise RunFailedError(desc.app_slug, desc.number)
+                    else:
+                        return False
 
-def wait_for_runs(runs: List[Run]) -> None:
+                elif _is_run_awaiting_completion(desc):
+                    time.sleep(WAIT_TIMEOUT)
```
datancoffee, line 218: @bradhe sleeping inside the single-run function is fine when the parent app is just waiting for 1 child app.

But when the parent is waiting for 2 or more apps, then this waiting here will block the status checks for the entire run group.

This is fine for v1, where we assume the best case and that users want all children to succeed before proceeding to the next step of the parent app.

But I would recommend in future versions to implement a heart beat of 2 seconds where we check for status of all children at the same time.

If one of the children fails, and the raise_on_error is true, then the parent will correctly terminate.

In the v1 implementation, the parent will keep executing the first child even if the second child already failed.

Not a big deal right now.

src/tower/_client.py
```diff
@@ -101,20 +110,63 @@ def run_app(
...
+        bool: True if the run is awaiting run or currently running, False otherwise.
+    """
+    return run.status in ["pending", "scheduled", "running"]
+
+
+def wait_for_run(
+    run: Run,
+    timeout: Optional[float] = 86_400.0, # one day
+    raise_on_failure: bool = False,
+) -> Run:
```
datancoffee, line 156: @bradhe See my comment about sleeping inside the wait_for_run vs wait_for_runs

Perhaps add a parameter "wait_for_completion: bool = True" which should be True when user waits for 1 run
But when this function is called by wait_for_runs, the parameter should be False, and the waiting should be done in wait_for_runs

src/tower/_client.py
bradhe: I've addressed this in the updated `wait_for_runs` implementation.

src/tower/_client.py
bradhe: I've separated `wait_for_run` and `wait_for_runs` entirely to deal with timeout semantics separately, so no worries here.

src/tower/_client.py
bradhe: Yes, this was an incomplete implementation. Test coverage now validates this, too.

datancoffee:
Solid piece of German engineering. Approved.

PR #49, opened 2025-05-13
Handle field type mappings correctly

sankroh:
This PR:
- fixes the arrow type to iceberg type mapping for fields with `timestamptz` types
- fixes the handling of ids and nested datatypes like Lists and Structs
- applies formatting to the code
- adds tests for the various type mappings


bradhe (approved):
This is awesome, great work! Super helpful.

PR #58, opened 2025-07-04
Release v0.3.20

bradhe:
- Upgrade the Tower API
- Move to rustls instead of native TLS
- Bump version to v0.3.20


konstantinoscs (approved):
Awesomeeee

PR #62, opened 2025-07-08
Add OpenAPI spec to repo

socksy:
This pulls the OpenAPI spec we generate the clients from into the git repo so that we have a snapshot in time of exactly which version it depended on. Useful for a contract test, which can go see what the latest client depends on.


bradhe:
Hmm on one side is YAML and the other side, uses JSON. Maybe we should use YAML everywhere? Not sure if the generator can take YAML directly?

socksy:
Ooh, I hadn't even noticed 🤦 

Could always save the JSON as YAML given it's supposed to be a superset... I'll have a look when I'm next working on this.

PR #66, opened 2025-07-16
Use `uv` to manage Python runtime and state

bradhe:
This PR migrates the python execution pipeline to use `uv` instead of raw Python everywhere. It has some code for making sure `uv` exists/runs on the relevant host, too.


crates/tower-runtime/src/local.rs
```diff
@@ -431,28 +351,6 @@ fn make_env_vars(ctx: &tower_telemetry::Context, env: &str, cwd: &PathBuf, is_vi
         res.insert(key.to_string(), value.to_string());
     }
 
-    // If we're in a virtual environment, we need to add the bin directory to the PATH so that we
-    // can find any executables that were installed there.
-    if is_virtualenv {
```
konstantinoscs, line 436: So uv takes care of this?

konstantinoscs (commented):
Can you run a `build-tower-runner-binaries` on this? I see changes to cargo.toml and I'm thinking that we should have some kind of way to safeguard against regressing from pure rustls...

PR #74, opened 2025-07-27
Initial check-in of changes to llms() API improving model name resolution

datancoffee:
The goals of this PR was to allow developers to use LLMs across their development and production environments just by specifying a short model family name and without having lots of conditional statements handling differences in environments.

E.g. calling llms("llama3.2") should resolve to the model "llama3.2:3b" when working with ollama local inference and to model "meta-llama/Llama-3.2-3B-Instruct" when using Hugging Face Hub serverless inference.

Differences in naming of models, and the fact that each model is actually a collection of dozen of versions that differ in number of parameters (distillation) and quantization techniques made the naming resolution a tricky task.

Here are the new capabilities of the llms() API:

1. Tower now recognizes ~170 names of model families as of August 2025. Btw, there are 100,000+ individual models in these model families

2. Users can specify a model family e.g. llms("deepseek-r1") in both local and Tower cloud environments, and Tower will resolve the model family to a particular model that is available for inference:
- in local environment, Tower will find the model that is installed and, if there are multiple installed, pick the one with the largest number of parameters
- in Tower cloud environments, Tower will take the first model returned by HF search, making sure that this model is servable by the Inference Service, if specified by users

In addition to using model family names, Users can also specify a particular model both in local and Tower cloud environments: 
- locally: llms("deepseek-r1:14b") or llms("llama3.2:latest") 
- serverlessly: llms("deepseek-ai/DeepSeek-R1-0528") or llms("meta-llama/Llama-3.2-3B-Instruct")

Expected use:

A developer wants to use use a model of the "llama3.2" family first in development and then in production. They would add this code to their Tower app:

```
model_name=os.getenv("MODEL_NAME")
llm = llms(model_name)
```

They would set up their environments as follows:

"dev" environment:

Tower Secrets:

MODEL_NAME = "llama3.2"
(any model of this family installed locally will do)

TOWER_INFERENCE_ROUTER=ollama

"prod" environment:

Tower Secrets:

MODEL_NAME = meta-llama/Llama-3.2-3B-Instruct
(use a particular model)

TOWER_INFERENCE_ROUTER=hugging_face_hub
TOWER_INFERENCE_ROUTER_API_KEY=hf_123456789
TOWER_INFERENCE_SERVICE=<together|...>

### DIFF
--- a/src/tower/_llms.py
+++ b/src/tower/_llms.py
@@ -3,78 +3,457 @@ from typing import List
 from ollama import chat, pull
 from ollama import ChatResponse
 from ollama import ResponseError
+from ollama import list as ollama_list_models
+#import psutil # TODO: add this back in when implementing memory checking for LLMs
 
 from huggingface_hub import InferenceClient, ChatCompletionOutput
+from huggingface_hub import HfApi
 
 from ._context import TowerContext
 
-"""
-OLLAMA_MODELS and HUGGING_FACE_MODELS are dictionaries that map published model
-names to the internal names used by Tower when routing LLM requests to the
-underlying provider.
-"""
-OLLAMA_MODELS = {
-    "deepseek-r1": "deepseek-r1:14b",
-}
+LOCAL_INFERENCE_ROUTERS = [
+    "ollama", 
+    "vllm",
+]
 
-HUGGING_FACE_MODELS = {
-    "deepseek-r1": "deepseek-ai/DeepSeek-R1",
-}
+INFERENCE_ROUTERS = LOCAL_INFERENCE_ROUTERS + [
+    "hugging_face_hub"
+]
 
-def extract_model_name(ctx: TowerContext, supported_model: str) -> str:
+RAW_MODEL_FAMILIES = [
+    "all-minilm",
+    "aya",
+    "aya-expanse",
+    "athene-v2",
+    "bakllava",
+    "bge-large",
+    "bge-m3",
+    "cogito",
+    "codegemma",
+    "codegeex4",
+    "codeqwen",
+    "codestral",
+    "codeup",
+    "codellama",
+    "command-a",
+    "command-r",
+    "command-r-plus",
+    "command-r7b",
+    "deepcoder",
+    "deepseek-coder",
+    "deepseek-coder-v2",
+    "deepseek-llm",
+    "deepseek-r1",
+    "deepseek-v2",
+    "deepseek-v2.5",
+    "deepseek-v3",
+    "deepscaler",
+    "devstral",
+    "dbrx",
+    "dolphin-mistral",
+    "dolphin-mixtral",
+    "dolphin-phi",
+    "dolphin3",
+    "dolphincoder",
+    "exaone-deep",
+    "exaone3.5",
+    "everythinglm",
+    "falcon",
+    "falcon3",
+    "gemma",
+    "gemma2",
+    "gemma3",
+    "gemma3n",
+    "glm4",
+    "goliath",
+    "granite-code",
+    "granite-embedding",
+    "granite3-dense",
+    "granite3-guardian",
+    "granite3-moe",
+    "granite3.1-dense",
+    "granite3.1-moe",
+    "granite3.2",
+    "granite3.2-vision",
+    "granite3.3",
+    "hermes3",
+    "internlm2",
+    "lafrican",
+    "llama-pro",
+    "llama-guard3",
+    "llama2",
+    "llama2-chinese",
+    "llama2-uncensored",
+    "llama3",
+    "llama3-chatqa",
+    "llama3-groq-tool-use",
+    "llama3-gradient",
+    "llama3.1",
+    "llama3.2",
+    "llama3.2-vision",
+    "llama3.3",
+    "llama4",
+    "llava",
+    "llava-llama3",
+    "llava-phi3",
+    "magicoder",
+    "magistral",
+    "marco-o1",
+    "mathstral",
+    "meditron",
+    "medllama2",
+    "megadolphin",
+    "minicpm-v",
+    "mistral",
+    "mistral-large",
+    "mistral-nemo",
+    "mistral-openorca",
+    "mistral-small",
+    "mistral-small3.1",
+    "mistral-small3.2",
+    "mistrallite",
+    "moondream",
+    "mxbai-embed-large",
+    "nemotron",
+    "nemotron-mini",
+    "neural-chat",
+    "nexusraven",
+    "notus",
+    "nous-hermes",
+    "nous-hermes2",
+    "nous-hermes2-mixtral",
+    "nomic-embed-text",
+    "notux",
+    "olmo2",
+    "opencoder",
+    "openchat",
+    "openthinker",
+    "openhermes",
+    "orca-mini",
+    "orca2",
+    "paraphrase-multilingual",
+    "phi",
+    "phi3",
+    "phi3.5",
+    "phi4",
+    "phi4-mini",
+    "phi4-mini-reasoning",
+    "phi4-reasoning",
+    "phind-codellama",
+    "qwen",
+    "qwen2",
+    "qwen2-math",
+    "qwen2.5",
+    "qwen2.5-coder",
+    "qwen2.5vl",
+    "qwen3",
+    "qwq",
+    "r1-1776",
+    "reader-lm",
+    "reflection",
+    "sailor2",
+    "samatha-mistral",
+    "shieldgemma",
+    "smallthinker",
+    "smollm",
+    "smollm2",
+    "snowflake-arctic-embed",
+    "snowflake-arctic-embed2",
+    "solar",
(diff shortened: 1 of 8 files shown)


src/tower/_llms.py
```diff
@@ -3,78 +3,469 @@
...
+        return requested_model
     else:
-        if supported_model not in HUGGING_FACE_MODELS:
-            raise ValueError(f"Model {supported_model} not supported for Hugging Face Hub.")
-        return HUGGING_FACE_MODELS[supported_model]
+        raise ValueError(f"Model {requested_model} is not available locally. Please pull it first using 'ollama pull {requested_model}'")
+
+def resolve_vllm_model_name(ctx: TowerContext, requested_model: str) -> str:
+    raise NotImplementedError("vLLM is not supported yet.")
+    return requested_model
```
bradhe, line 317: This `return` is not reachable!

src/tower/_llms.py
```diff
@@ -3,78 +3,469 @@
...
+    api = HfApi(token=ctx.inference_router_api_key)
+
+    models = []
+
+    try:
+        model_info = api.model_info(requested_model, expand="inferenceProviderMapping")
+        models = [model_info]
+    except Exception as e:
+        # If model_info fails, fall back to search
+        pass
```
bradhe, line 333: Can you make this a bit more specific? Catching `Exception` will wallpaper over a whole bunch of different types of errors.

src/tower/_llms.py
```diff
@@ -3,78 +3,469 @@
...
+                'inference_providers': model.inference_provider_mapping
+            }
+            
+            # If inference_service is specified, only add models that support it
+            if ctx.inference_service is not None:
+                if ctx.inference_service not in [mapping.provider for mapping in model.inference_provider_mapping]:
+                    continue
+            
+            # Check that requested_model is partially contained in model.id
+            if normalize_model_family(requested_model) not in normalize_model_family(model.id):
```
bradhe, line 368: Wonder why this is `not in` and not an equality check?

src/tower/_llms.py
```diff
@@ -125,3 +519,28 @@ def infer_with_hugging_face_hub(ctx: TowerContext, model: str, messages: List, *
...
+#     try:
+#         memory = psutil.virtual_memory()
+#         return {
+#             'total': memory.total,
+#             'available': memory.available,
+#             'used': memory.used,
+#             'percent': memory.percent
+#         }
+#     except Exception as e:
+#         raise RuntimeError(f"Failed to get memory information: {str(e)}")
```
bradhe, line 544: Don't think we need this any longer, ya?

src/tower/_llms.py
```diff
@@ -3,78 +3,469 @@
 from ollama import chat, pull
 from ollama import ChatResponse
 from ollama import ResponseError
+from ollama import list as ollama_list_models
+#import psutil # TODO: add this back in when implementing memory checking for LLMs
```
bradhe, line 7: Also don't think we need this any longer?

tests/tower/test_llms.py
```diff
@@ -0,0 +1,219 @@
...
+            
+            # Verify it's an Llm instance
+            assert isinstance(llm, Llm)
+
+            # Verify the context was set
+            assert llm.context == mock_hf_together_context
+
+            # Test a simple prompt
+            response = llm.prompt("Hello, how are you?")
+            assert response == "This is a test response"
```
bradhe, line 166: I can't quite follow, but this doesn't require that the test actually talks to HuggingFace, correct? It is mocked using the `mock_complretion` above if I understand this correctly, yes?

tests/tower/test_llms.py
```diff
@@ -0,0 +1,219 @@
...
+            
+            # Verify the resolved model was found on the Hub
+            assert llm.model_name.startswith("deepseek-ai/DeepSeek-R1")
+
+
+def test_llms_inference_with_hugging_face_hub_1(mock_hf_together_context):
+    """Test actual inference on a model served by together via Hugging Face Hub."""
+    # Mock the TowerContext.build() to return our mock context
+    with patch('tower._llms.TowerContext.build', return_value=mock_hf_together_context):
```
bradhe, line 175: Maybe we need to provide a way of injecting expected context into the environment. Like, there should be a `with_context` function in the `tower` that allows us to set context? Would allow us to avoid this hairy `patch` calls everywhere.

src/tower/_llms.py
```diff
@@ -3,78 +3,469 @@
...
+
+        # Check that we know this router. This will also check that router was set when not in local mode.
+        if context.inference_router not in INFERENCE_ROUTERS:
+            raise ValueError(f"Inference router {context.inference_router} not supported.")
+        
+        self.model_name = resolve_model_name(
+            self.context, self.requested_model_name)
+        
+
+    def chat_completion(self, messages: List) -> str:
```
bradhe, line 409: This should be a verb, to match the rest of the conventions in here! Maybe `complete`?

bradhe:
Also can you do a `uv sync`? Looks like the `uv.lock` is out of date!

tests/tower/test_llms.py
datancoffee: Yes, you are correct. Claude was too smart for its own good. 
I simplified this test to just test the name resolution of a model, with no fake inference involved

datancoffee:
> 2\. "inference router" what VLLM and HuggingFace call themselves? I'm good as long as we're not inventing it!

As discussed, I will rename inference_service to inference_provider, because this is terminology used by hugging face.

Inference Router is something I came up with, because we are dealing with services that route inference requests to inference providers.

src/tower/_llms.py
datancoffee: Here is an example of how HF Hub does it:

from huggingface_hub import InferenceClient

client = InferenceClient("meta-llama/Meta-Llama-3-8B-Instruct")

# Perform a chat completion
response = client.chat_completion(
    messages=[{"role": "user", "content": "What is the capital of France?"}],
)

src/tower/_llms.py
```diff
@@ -3,78 +3,469 @@
...
+            # TODO: add this back in when implementing memory checking for LLMs
+            # raise ValueError(f"No models in family {requested_model} fit in available memory ({memory['available'] / (1024**3):.2f} GB) with max memory threshold {MEMORY_THRESHOLD} or are not available locally. Please pull a model first using 'ollama pull {requested_model}'")
+            raise ValueError(f"No models in family {requested_model} are available locally. Please pull a model first using 'ollama pull {requested_model}'")
+    elif requested_model in local_model_names:
+        return requested_model
     else:
-        if supported_model not in HUGGING_FACE_MODELS:
-            raise ValueError(f"Model {supported_model} not supported for Hugging Face Hub.")
-        return HUGGING_FACE_MODELS[supported_model]
+        raise ValueError(f"Model {requested_model} is not available locally. Please pull it first using 'ollama pull {requested_model}'")
```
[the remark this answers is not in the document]
datancoffee: @bradhe 
yeah, i saw you had that code in there previously. The rationale for letting the user install the model themselves is that these models are anywhere between 1GB and 100GB (or even more) both on disk and in memory. Since ollama and our app are running on the same node, it is too easy to shoot yourself in the foot and download a 100GB model by mistake. That's why I modified model name resolution logic to only look for already installed models. But I left your code in the "chat_completion_with_ollama" function. In the future we might do something smarter with downloads

src/tower/_llms.py
datancoffee: @bradhe I removed the reference to vllm for now, because we don't really integrate with it now. Will add back in the future, and will most probably use the HuggingFace local client, as you suggested

src/tower/_llms.py
bradhe: Right I meant our own conventions--we use verbs for all our function names.

src/tower/_llms.py
datancoffee: Changed it to differentiate between non-existent name and other errors

try:
        model_info = api.model_info(requested_model, expand="inferenceProviderMapping")
        models = [model_info]
    except RepositoryNotFoundError as e:
        # If model_info fails, it means the model does not exist under this exact name
        # Therefore, fall back to "search" and look for models that partially match the name
        # In Hugging Face Hub terminology Repository = Model / Dataset / Space.
        pass
    except Exception as e:
        # for the rest of the errors, we will raise an error
        raise RuntimeError(f"Error while getting model_info for {requested_model}: {str(e)}")

tests/tower/test_llms.py
datancoffee: If you don't mind, I won't do this in this PR, as I am not sure how to do it.

src/tower/_llms.py
datancoffee: Changed to complete_chat(), as discussed

bradhe (approved):
Disable the failing tests and you can land this. We need a green build, because we need to cut a release such that we can build you a binary, etc., to install.

PR #80, opened 2025-08-19
fix: don't assume 422 are validation errors

konstantinoscs:
Until now, all UNPROCESSABLE_ENTITY errors were assumed to be validation errors. With quotas, that's no longer true.


bradhe (approved):
Discussed on a Hangout and this approach will work well for now!

PR #86, opened 2025-09-01
MCP Server

socksy:
This PR introduces an SSE MCP server that you can run from any tower project (or wannabe tower project) with `tower mcp-server`). It provides the following tools (in order as they appear in claude, at least, which is unfortunately somewhat illogical):

1. list all your apps
2. get help on how the workflow works (a tool that prints out documentation for the LLM to understand what to do — not sure if this is a standard solution but seemed quite effective in getting it to do the right thing)
3. create secrets
4. deploy app
5. add a parameter to an existing Towerfile
6. switch teams
7. validate a Towerfile
8. run a tower app (on our servers, so "remote")
9. run a tower app (on your computer, so "local")
10. look at the logs of a run
11. generate a Towerfile from a pyproject.toml — added this functionality to a new rust file, if it works well enough we should consider integrating it into our tool, either as a manual command or a transparent step. Probably needs some more testing but it worked on a couple small examples I gave it
12. read/parse an existing Towerfile
13. updates a Towerfile
14. delete secrets
15. list teams
16. show details of an app and its recent runs
17. list all secrets (the names + a small preview of the actual secret)
18. create an app

This PR is exactly the same as what's in #81, but I've got claude to dramatically reduce the amount of commits and to reword the commit messages. I hope that makes it easier to follow — but in case it makes something a bit illogical, the true history is [here](https://github.com/tower/tower-cli/pull/81/commits). Since it was a lot of exploring, there's a lot of stuff that was added in some commits, and the removed/reworked completely in later commits.

I will write proper documentation for this MCP server later, but for now it would be nice to get this in.


crates/config/src/towerfile.rs
```diff
@@ -1,8 +1,8 @@
 use crate::Error;
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
```
socksy, line 2: now we can both read and write towerfiles, we add the Serialize trait to everything that previously had Deserialize

crates/tower-cmd/src/towerfile_gen.rs
```diff
@@ -0,0 +1,271 @@
...
+    fn find_any_python_file(dir: &Path) -> Option<String> {
+        Self::get_python_files(dir)
+            .into_iter()
+            .next()
+            .map(|name| format!("./{}", name))
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
socksy, line 116: unit tests are inline in rust

crates/tower-cmd/src/mcp.rs
```diff
@@ -0,0 +1,512 @@
...
+    
+    fn resolve_working_directory(common: &CommonParams) -> std::path::PathBuf {
+        common.working_directory
+            .as_ref()
+            .map(std::path::PathBuf::from)
+            .unwrap_or_else(|| std::env::current_dir().unwrap_or_else(|_| std::path::PathBuf::from(".")))
+    }
+
+
+    #[tool(description = "List all Tower apps in your account")]
```
socksy, line 160: The rust MCP library, `rmcp`, has this nice abstraction where you annotate async functions (in a `tool_router` struct implementation) with the description of the tool. It then picks out them all and generates the endpoints and everything needed for the MCP server.

crates/tower-cmd/src/mcp.rs
```diff
@@ -0,0 +1,512 @@
...
+        match std::fs::write(&towerfile_path, &content) {
+            Ok(_) => {
+                let success_msg = format!("Generated Towerfile at {}\n\n{}", towerfile_path.display(), content);
+                Self::text_success(success_msg)
+            }
+            Err(e) => Self::error_result("Failed to write Towerfile", e),
+        }
+    }
+
+    #[tool(description = "Show the recommended workflow for developing and deploying Tower applications")]
```
socksy, line 453: No idea if other MCP tools do this, but I wanted to find a way to give the LLM an ability to learn more about how to use the system. If it were all in the docs directly, then all that text would be prepended to every single prompt for any user using claude code, which seems unideal, since they won't always be working on tower stuff. This way we can have just the docstring for each thing, and then it can choose to call this tool to dynamically learn how to use all the other tools (and specifically, what order it should use them in). 

From my testing, it seems to work well, with claude often (always?) choosing to call this tool for docs first.

README.md
```diff
@@ -147,6 +147,144 @@ pprint.pprint(tower.get_available_features())
...
+      "url": "http://127.0.0.1:34567/sse",
+      "transport": "sse"
+    }
+  }
+}
+```
+
+For custom ports, adjust the URL accordingly (e.g., `http://127.0.0.1:8080/sse`).
+
+##### Cursor
```
socksy, line 214: I don't use Cursor or Windsurf so I've not tested either those integrations... Maybe if someone does it would be nice to double check if this setup is correct.

crates/tower-cmd/src/run.rs
```diff
@@ -59,9 +60,16 @@ pub fn run_cmd() -> Command {
         .about("Run your code in Tower or locally")
 }
 
+pub async fn do_run(config: Config, args: &ArgMatches, cmd: Option<(&str, &ArgMatches)>) {
```
socksy, line 63: this was a leftover from when I was calling the inner function from the mcp.rs, which I don't do anymore. The idea was that it separates the IO (`eprintln!` and exiting) from the pure part of actually running and reading from a stream. I could probably remove it now, especially since `do_run_inner` isn't an amazing name. However I do like splitting side effectful IO into a wrapper function in general so I didn't feel the urge.

crates/tower-cmd/src/run.rs
```diff
@@ -138,65 +144,99 @@ async fn do_run_local(config: Config, path: PathBuf, env: &str, mut params: Hash
...
+/// the package, and launch the app. The relevant package is cleaned up after execution is
+/// complete.
+pub async fn do_run_local(config: Config, path: PathBuf, env: &str, params: HashMap<String, String>) -> Result<()> {
+    do_run_local_impl(config, path, env, params, |receiver| async {
+        monitor_output(receiver).await;
+        ()
+    }).await
+}
+
+/// do_run_local_capture is like do_run_local but returns captured output lines instead of printing
```
socksy, line 185: we need to have something like this in order to redirect the streamed local output to the mcp server instead of just printing it out (which in this case would mean that the terminal window you run `tower mcp-server` from would get the logs, but the LLM wouldn't be able to see them)

crates/tower-cmd/src/run.rs
```diff
@@ -216,86 +256,146 @@ async fn do_run_remote(
...
 
+async fn do_follow_run_impl(
+    config: Config,
+    run: &Run,
+    sink: &dyn OutputSink,
+    enable_ctrl_c: bool,
+) -> Result<()> {
     match wait_for_run_start(&config, &run).await {
         Err(err) => {
-            spinner.failure();
```
socksy, line 231: we really don't want this in the main output, because then it would generate a thousand spinning braille events that would be parsed as separate lines in the mcp server.

Note to self: I need to double check it's still there when running directly in the cli 🙃

crates/tower-cmd/src/run.rs
```diff
@@ -216,86 +256,146 @@ async fn do_run_remote(
...
 ) {
-    let mut spinner = output::spinner("Waiting for run to start...");
+    let sink = StdoutSink;
+    let _ = do_follow_run_impl(config, run, &sink, true).await;
+}
 
+async fn do_follow_run_impl(
+    config: Config,
+    run: &Run,
+    sink: &dyn OutputSink,
```
socksy, line 317: so we use tokio channels for streaming the output now. You can see through this code we now have an OutputSink and StdoutSink — a specific implementation of the OutputSink channel that writes to standard out. This way we can use another channel to get the run output in the MCP server, and stream it directly to the client using SSE events (hence using an SSE server, which sort of seems to be the most common type of MCP server anyway?)

crates/tower-cmd/src/run.rs
```diff
@@ -374,114 +474,149 @@ fn get_app_name(cmd: Option<(&str, &ArgMatches)>) -> Option<String> {
...
+    Ok(vals)
 }
 
 /// load_towerfile manages the process of loading a Towerfile from a given path in an interactive
 /// way. That means it will not return if the Towerfile can't be loaded and instead will publish an
 /// error.
-fn load_towerfile(path: &PathBuf) -> Towerfile {
-    Towerfile::from_path(path.clone()).unwrap_or_else(|err| {
+fn load_towerfile(path: &PathBuf) -> Result<Towerfile> {
+    Towerfile::from_path(path.clone()).map_err(|err| {
```
socksy, line 519: Before, `unwrap_or_else` tries to take the result of a function that returns Result<Self, Error>, and replaces the error result with this new value... which in this case just called `std::process::exit(1)` (so I guess didn't replace it?). Now we want to keep the Result type, so we want to specifically call a function when we get an error response so that we can add some debug info, which is what `map_err` does in rust (i.e. it "maps" a function to an "err" value... personally I would have named this function something else but it's a pretty well known rust idiom by now).

tests/integration/templates/Towerfile.j2
```diff
@@ -0,0 +1,8 @@
+[app]
```
socksy, line 1: this ends with `.j2` as if it's a jinja template, and it sort of is, but in this case I'm quite literally doing a find and replace for these specific variables: https://github.com/tower/tower-cli/pull/86/commits/6f04369223e9dd3739f6819900edea03ac5014cf#diff-365b24b9adc19a6d5ff90a9eb58bbc764d70284234d4dcc97bd35bc5d5a7b5f3R43

crates/tower-cmd/src/towerfile_gen.rs
```diff
@@ -0,0 +1,271 @@
...
+        Ok(toml::to_string(&towerfile)?)
+    }
+
+
+    fn find_main_script(dir: &Path) -> Option<String> {
+        Self::find_script_from_pyproject(dir)
+            .or_else(|| Self::find_script_with_main(dir))
+            .or_else(|| Self::find_common_script(dir))
+            .or_else(|| Self::find_any_python_file(dir))
+    }
```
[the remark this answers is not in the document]
socksy: You mean other than the script, as taken on line 55?

crates/tower-cmd/Cargo.toml
```diff
@@ -4,7 +4,6 @@ version = { workspace = true }
 edition = "2021"
 
 [dependencies]
-anyhow = { workspace = true }
```
bradhe, line 7: 🙌

bradhe (approved):
LGTM!

PR #90, opened 2025-09-07
Remove async code from the Drop handler

bradhe:
Since the code was async, `terminate` was never actually called. In order to avoid all this crazy, we just push all the async code out (as much as possible) to make things simpler. We don't explicitly need to join the execution handler after the child is terminated...it just seems like the right thing to do to ensure the Drop implementation waits for the whole thing to be dead.

### DIFF
--- a/crates/tower-runtime/src/local.rs
+++ b/crates/tower-runtime/src/local.rs
@@ -17,6 +17,7 @@ use tokio::{
     fs,
     io::{AsyncRead, BufReader, AsyncBufReadExt},
     process::{Child, Command}, 
+    runtime::Handle,
     sync::{
         Mutex,
         oneshot::{
@@ -49,7 +50,7 @@ pub struct LocalApp {
     waiter: Mutex<oneshot::Receiver<i32>>,
 
     // terminator is what we use to flag that we want to terminate the child process.
-    terminator: Mutex<CancellationToken>,
+    terminator: CancellationToken,
 
     // execute_handle keeps track of the current state of the execution lifecycle.
     execute_handle: Option<JoinHandle<Result<(), Error>>>,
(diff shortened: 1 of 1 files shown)


crates/tower-runtime/src/local.rs
```diff
@@ -271,20 +272,28 @@ async fn execute_local_app(opts: StartOptions, sx: oneshot::Sender<i32>, cancel_
...
+        // CancellationToken::cancel() is not async
+        self.terminator.cancel();
+        
+        // Optionally spawn a task to wait for the handle
+        if let Some(execute_handle) = self.execute_handle.take() {
+            if let Ok(handle) = Handle::try_current() {
+                handle.spawn(async move {
+                    let _ = execute_handle.await;
+                });
+            }
```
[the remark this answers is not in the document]
bradhe: This is intentional. It's not strictly required and we don't care if it fails really.

crates/tower-runtime/src/local.rs
```diff
@@ -271,20 +272,28 @@ async fn execute_local_app(opts: StartOptions, sx: oneshot::Sender<i32>, cancel_
...
-        let _ = self.terminate();
+        // CancellationToken::cancel() is not async
+        self.terminator.cancel();
+        
+        // Optionally spawn a task to wait for the handle
+        if let Some(execute_handle) = self.execute_handle.take() {
+            if let Ok(handle) = Handle::try_current() {
+                handle.spawn(async move {
+                    let _ = execute_handle.await;
+                });
```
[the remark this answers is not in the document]
bradhe: Same as above.

crates/tower-runtime/src/local.rs
```diff
@@ -421,11 +438,57 @@ fn make_env_vars(ctx: &tower_telemetry::Context, env: &str, cwd: &PathBuf, secs:
...
+    // deadline).
+    let timeout = timeout(
+        Duration::from_secs(2), 
+        child.wait()
+    ).await;
+    
+    if timeout.is_err() {
+        killpg(
+            pid,
+            Signal::SIGKILL
```
socksy, line 474: maybe a bit old school but i reckon a SIGHUP before SIGKILL might be an idea

socksy (approved):

crates/tower-runtime/src/local.rs
```diff
@@ -421,11 +438,57 @@ fn make_env_vars(ctx: &tower_telemetry::Context, env: &str, cwd: &PathBuf, secs:
     res
 }
 
+#[cfg(unix)]
+async fn kill_child_process(ctx: &tower_telemetry::Context, mut child: Child) {
+    let pid = match child.id() {
+        Some(pid) => pid,
+        None => {
+            // We didn't get anything, so we can't do anything. Let's just exist with a debug
```
konstantinoscs, line 446:
```suggestion
            // We didn't get anything, so we can't do anything. Let's just exit with a debug
```
?

crates/tower-runtime/src/local.rs
```diff
@@ -421,11 +438,57 @@ fn make_env_vars(ctx: &tower_telemetry::Context, env: &str, cwd: &PathBuf, secs:
...
+    if timeout.is_err() {
+        killpg(
+            pid,
+            Signal::SIGKILL
+        ).ok();
+    }
+}
+
+#[cfg(not(unix))]
+async fn kill_child_process(ctx: &tower_telemetry::Context, mut child: Child) {
```
konstantinoscs, line 480: Ok so in windows will default back to whatever `child.kill()` tries to do

konstantinoscs (approved):
In general LGTM

crates/tower-runtime/src/local.rs
bradhe: Yes, apparently (untested by me) it's not an issue on Windows.

PR #93, opened 2025-09-10
Update clients

konstantinoscs:
Pretty straight-forward


konstantinoscs:
Actually I'll amend this because we're about to update our API spec server-side

sammuti (approved):
LGTM, not sure if my approval counts as much 😅

PR #99, opened 2025-09-16
Adding drop table feature for tower tables

sammuti:


src/tower/_tables.py
```diff
@@ -487,6 +487,43 @@ def create_if_not_exists(self, schema: pa.Schema) -> Table:
...
+        table_name = make_table_name(self._name, namespace)
+
+        try:
+            self._catalog.drop_table(table_name)
+            return True
+        except Exception:
+            # If the table doesn't exist or there's any other issue, return False
+            # The underlying PyIceberg catalog will raise different exceptions
+            # depending on the catalog implementation, so we catch all exceptions
+            return False
```
bradhe, line 524: Hmm do we want to swallow all exceptions here? Can we find the table does not exist exception in particular, and throw the other ones?

bradhe (changes requested):
One quick discussion point!

PR #101, opened 2025-09-17
Introducing black formatting for python code

sammuti:
- Introduce black pre-commit hook
- Format existing code


socksy (approved):
lookin good! I know it's in draft but already an improvement. If you add a --check command to one of the github workflows (maybe `.github/workflows/test-python.yml`?) and a line in the README for how to run it, that'd be grand

sammuti:
mistake, I need to make a change so that it will be available to you locally

PR #102, opened 2025-09-19
Fix run regressions

socksy:
The MCP server resulted in a bunch of regression for running an app via the actual CLI — this was because I refactored it to have a pure inner function and a side-effectful outer function, and re-used the inner function in both the MCP server and in the CLI. I tried to keep most of the functionality by using a ChannelSink at the time, with the idea that we could still stream logs in. However, fundamentally the behaviour needs to be different.

So:
- remote runs have newlines again (😅)
- there's two different sinks now, `ChannelSink` for the normal formatted lines, and `PlainChannelSink` for no formatting (for MCP server responses)
- 422 errors were not being successfully interpreted as errors by the MCP server, and fundamentally it seemed to be because the `unwrap_api_response` from the api.rs didn't treat them as errors (which they're not really — they're 4xx issues). I'm not 100% sure this solution is the best and am interested in other suggestions

  However we want to be able to see we've passed in the wrong parameters to a run, and the MCP server should accordingly return with an error when this fails, so now we wrap the call to `unwrap_api_response` with a check to see if there was a client error or server error — and when there is, return an actual Err response that we can match for in the sink logic (see [here](https://github.com/tower/tower-cli/compare/develop...fix/cli-run-regressions?expand=1#diff-7d1ee340d973bf920db1c89e6bf40a06603127b382cc4bf345558b9fe528f4dbR217) and [here](https://github.com/tower/tower-cli/compare/develop...fix/cli-run-regressions?expand=1#diff-7d1ee340d973bf920db1c89e6bf40a06603127b382cc4bf345558b9fe528f4dbR245))
- reintroduce spinners (and pass in a boolean to enable or disable them so as not to spam the MCP server)
- timestamps are colourized again
- the lines appear one by one. Not sure exactly why that was broken before, maybe the fact it was all on one line they got buffered
- local runs correctly detect when the app errored out locally (this got subtly broke in the refactor, `status_task.abort();` got called before we printed out the failure)
- lots of regression tests so this won't happen again (and of course, the tests caught a bunch of ways my fixes weren't working 🙃)

### DIFF
--- a/crates/tower-cmd/src/api.rs
+++ b/crates/tower-cmd/src/api.rs
@@ -1,13 +1,14 @@
+use std::collections::HashMap;
+
 use config::Config;
 use futures_util::StreamExt;
 use http::StatusCode;
 use reqwest_eventsource::{Event, EventSource};
-use std::collections::HashMap;
 use tokio::sync::mpsc;
-use tower_api::apis::configuration;
-use tower_api::apis::Error;
-use tower_api::apis::ResponseContent;
-use tower_api::models::RunParameter;
+use tower_api::{
+    apis::{configuration, Error, ResponseContent},
+    models::RunParameter,
+};
 use tower_telemetry::debug;
 
 /// Helper trait to extract the successful response data from API responses
--- a/crates/tower-cmd/src/api.rs
+++ b/crates/tower-cmd/src/api.rs
@@ -730,7 +730,6 @@ impl ResponseEntity for tower_api::apis::default_api::ListEnvironmentsSuccess {
     }
 }
 
-
 pub async fn list_environments(
     config: &Config,
 ) -> Result<
--- a/crates/tower-cmd/src/api.rs
+++ b/crates/tower-cmd/src/api.rs
@@ -156,7 +156,17 @@ pub async fn run_app(
         },
     };
 
-    unwrap_api_response(tower_api::apis::default_api::run_app(api_config, params)).await
+    match tower_api::apis::default_api::run_app(api_config, params).await {
+        Ok(response) if response.status.is_client_error() || response.status.is_server_error() => {
+            Err(Error::ResponseError(tower_api::apis::ResponseContent {
+                tower_trace_id: response.tower_trace_id,
+                status: response.status,
+                content: response.content,
+                entity: None,
+            }))
+        }
+        response => unwrap_api_response(async { response }).await,
+    }
 }
 
 pub async fn export_secrets(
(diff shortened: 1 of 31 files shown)


crates/tower-cmd/src/api.rs
```diff
@@ -156,7 +156,17 @@ pub async fn run_app(
...
-    unwrap_api_response(tower_api::apis::default_api::run_app(api_config, params)).await
+    match tower_api::apis::default_api::run_app(api_config, params).await {
+        Ok(response) if response.status.is_client_error() || response.status.is_server_error() => {
+            Err(Error::ResponseError(tower_api::apis::ResponseContent {
+                tower_trace_id: response.tower_trace_id,
+                status: response.status,
+                content: response.content,
+                entity: None,
+            }))
+        }
```
bradhe, line 167: Wonder if this can be generalized and, as Copilot suggests, this can get pushed into `update_api_response`? Nothing really fancy going on here.

crates/tower-cmd/src/mcp.rs
```diff
@@ -103,6 +103,13 @@ struct EmptyRequest {
     common: CommonParams,
 }
 
+#[derive(Debug, Deserialize, JsonSchema)]
+struct RunRequest {
+    #[serde(flatten)]
+    common: CommonParams,
+    parameters: Option<std::collections::HashMap<String, String>>,
```
bradhe, line 110: nit: I was just thinking that it'd be good to push `use std::collections::HashMap` onto the top of the file. But, then I thought there's probably a guideline for when we should `use` something versus directly referencing it, and I'd bet it's about how many times the thing is used in a file?

So, now I'm not sure :) Just thinking out loud, no real action in this comment but perhaps it will inspire conversation elsewhere.

crates/tower-cmd/src/run.rs
```diff
@@ -291,44 +327,91 @@ impl OutputSink for StdoutSink {
 struct ChannelSink(tokio::sync::mpsc::UnboundedSender<String>);
 impl OutputSink for ChannelSink {
     fn send_line(&self, line: String) {
-        let _ = self.0.send(line);
+        if let Some((ts, msg)) = line.split_once(": ") {
```
bradhe, line 330: Man this really makes me think that we should be pushing the timestamp into a parameter to `send_line` on the `OutputSink` trait, considering we're doing it in two places. Probably want to be able to format the ts differently for different cases...

Could be follow-up.

bradhe (approved):
Personally, I don't think there's anything critical here that should block the PR. I think there's a bit of follow-up/cleanup we should do later on!

socksy:
closed in favour of the PR stack that was eventually folded into fix/tow-963-switch-to-global-output-writing (#105)

PR #105, opened 2025-09-26
fix:use global output::write() function instead of sinks and conditional logic everywhere

socksy:
Pulled out of #102 in the interest of making things easier to review. Downside is that I'm not sure if the tests will pass on their own without all the other branches


bradhe (approved):
I think this is a good iteration.

crates/tower-cmd/src/output.rs
```diff
@@ -14,6 +17,29 @@ use tower_telemetry::debug;
...
 
+static CAPTURE_MODE: OnceLock<bool> = OnceLock::new();
+static CURRENT_SENDER: Mutex<Option<UnboundedSender<String>>> = Mutex::new(None);
+
+pub fn set_capture_mode() {
+    CAPTURE_MODE.set(true).ok();
+}
+
+pub fn set_current_sender(sender: UnboundedSender<String>) {
+    *CURRENT_SENDER.lock().unwrap() = Some(sender);
```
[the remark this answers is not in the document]
socksy: I don't think this is a meaningful error message for an end user, and I'm struggling to come up with an error message that would at all be helpful to them. If we're going with unhelpful error messages, I don't see why we should bother changing from the
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: PoisonError { .. }', src/output.rs:32:5
```
that we would expect from this one

crates/tower-cmd/src/deploy.rs
```diff
@@ -29,61 +35,69 @@ fn resolve_path(args: &ArgMatches) -> PathBuf {
...
 
 pub async fn do_deploy(config: Config, args: &ArgMatches) {
-    // Determine the directory to build the package from
     let dir = resolve_path(args);
+    let auto_create = args.get_flag("auto-create");
+    if let Err(err) = deploy_from_dir(config, dir, auto_create).await {
+        match err {
+            crate::Error::ApiDeployError { source } => output::tower_error(source),
+            crate::Error::ApiDescribeAppError { source } => output::tower_error(source),
+            _ => eprintln!("Deploy error: {}", err),
```
bradhe, line 44: Is there actually a different type of error that can be returned here? We should quickly fix this in a follow-up PR.

crates/tower-cmd/src/deploy.rs
```diff
@@ -29,61 +35,69 @@ fn resolve_path(args: &ArgMatches) -> PathBuf {
...
-                }
-            }
-        }
+    let package = match Package::build(spec).await {
+        Ok(package) => package,
         Err(err) => {
-            output::config_error(err);
+            spinner.failure();
+            let error = crate::Error::PackageError { source: err };
+            eprintln!("Package build failed: {}", error);
```
bradhe, line 78: This also shouldn't be here.

crates/tower-cmd/src/deploy.rs
socksy: well hopefully not, but when i was trying to debug it, yes 😬

PR #120, opened 2025-10-20
TOW-771 - logs being truncated 

socksy:
Looks like the server never closes the SSE stream, so it's up to us in the client (backend change forthcoming...). Presumably not a big deal on the web frontend because it's just constantly open, for now I'm adding a 5 second timeout after we detect the run is complete which should be plenty of time to get any remaining logs, and hopefully not so long that it makes people annoyed that it takes too long for the CLI to be responsive after setting a run in motion.


sammuti (approved):
LGTM

bradhe (approved):
LGTM, I think this is a decent solution. There's a lot of complexity in here that could be solved with the control plane issuing a EOF message 😄

PR #126, opened 2025-11-03
TOW-1130 - clean up deps to be optional/dev when not needed by core

socksy:
Resolves TOW-1130

Not sure how we ended up with ollama, hugging face and iceberg duplicated in the core dependencies, but we don't need them there, since they already have their own dependency groups. 

MCP is only needed for the integration test of the Rust CLI, so absolutely no need to put that in the main dependencies. 

Python 3.10 was needed by the MCP library, but it's already there in the dev dependencies next to MCP for that very reason.


pyproject.toml
```diff
@@ -111,5 +107,5 @@ dev = [
   "openapi-python-client==0.24.3",
   "pre-commit==4.0.1",
   "pyiceberg[sql-sqlite]==0.9.1",
-  "tower[iceberg]",
```
konstantinoscs: Why this?

socksy: See 9a0cfcb09cde8df8afc05ed4b509509227b6bcab

PR #127, opened 2025-11-03
TOW-1131 - add py.typed to python sdk

socksy:
Resolves TOW-1131

See title :) No more needed to be done since there's already types, this just lets tooling know that there are types defined in the library itself it can use. See [PEP 561](https://peps.python.org/pep-0561/)


konstantinoscs (approved):

bradhe (approved):
Great!

PR #131, opened 2025-11-17
Local apps shouldn't time out in terms of running

bradhe:
This PR changes the timeout logic slightly for local apps to prevent them from "timing out" after some amount of time. They should be able to run for as long as they need to run!

I also changed the output logic to defer printing the success or error message strictly at the end of the run output.

### DIFF
--- a/tests/integration/features/cli_runs.feature
+++ b/tests/integration/features/cli_runs.feature
@@ -13,7 +13,7 @@ Feature: CLI Run Commands
     When I run "tower run --local" via CLI
     Then timestamps should be green colored
     And each log line should be on a separate line
-    And the final status should show "Your app crashed!" in red
+    And the final status should show "Your local run crashed!" in red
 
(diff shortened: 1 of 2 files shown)


crates/tower-cmd/src/run.rs
```diff
@@ -536,59 +545,54 @@ async fn monitor_output(mut output: OutputReceiver) {
...
-                if check_count >= max_checks {
-                    debug!(
-                        "Failed to get app status after {} attempts, giving up",
-                        max_checks
-                    );
-                    output::error("Failed to get app status after timeout");
-                    return tower_runtime::Status::Running; // Return a default status for timeout
+                err_count += 1;
+
+                // If we get five errors in a row, we abandon monitoring.
```
jo-sm: I don't see that we reset `err_count` anywhere, should we reset it if we get a non error?

PR #133, opened 2025-11-19
Pass the current run ID as the parent_run_id when calling `run_app`

jo-sm:
This PR gets the current run ID from the context and passes it to the API when calling `run_app`.


socksy (approved):
Do you need to also do this for the rust CLI?

PR #139, opened 2025-12-11
[TOW-1213] Remove the package information from the logs

sammuti:
Remove the package information from the logs using `uv --quiet ..`

### DIFF
--- a/crates/tower-uv/src/lib.rs
+++ b/crates/tower-uv/src/lib.rs
@@ -161,6 +161,7 @@ impl Uv {
                 .arg("--color")
                 .arg("never")
                 .arg("--no-progress")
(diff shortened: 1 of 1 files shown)


crates/tower-uv/src/lib.rs
```diff
@@ -161,6 +161,7 @@ impl Uv {
                 .arg("--color")
                 .arg("never")
                 .arg("--no-progress")
+                .arg("--quiet")
```
sammuti: Probably best to have some sort of `VERBOSE` or `DEBUG` flag to enable these since this may hide legit relevant problems sometimes... 
@bradhe any thoughts?

PR #141, opened 2025-12-16
[TOW-1213] Remove rust package info from app runtime log lines and No color for non-tty

sammuti:
Remove rust package info from app runtime log lines and No color for non-tty


bradhe (approved):
LGTM :)

PR #145, opened 2025-12-24
Always read jwt

socksy:
Changing team on the MCP server is broken, because we set the session JWT once on program startup. API calls now always read the session from disk. When using TOWER_JWT, the session is saved to disk with the correct active team based on the JWT's account ID.


bradhe (approved):
LGTM

PR #148, opened 2025-12-30
[TOW-1316] Validate 50MB Limit for deployment package size

sammuti:
Validate 50MB Limit for deployment package size


bradhe (approved):
Looks good overall, one minor change to messaging then you can land it!

PR #157, opened 2026-01-08
Better version comparison

socksy:
See title. Got annoyed with it always telling me I had to update my CLI when my CLI is actually newer than what it was telling me to update to.

### AI statement
Claude generated test, reasoned with claude re: parsing and manually modifying the local version back and forth to test it.


bradhe (approved):
lgtm

PR #160, opened 2026-01-13
Add some commonly-used environment variables to child app environments

bradhe:
Some users reported weird behavior when installing certain third-party libraries and using them with Tower. Turns out, they expected to have a `$HOME` env var to save some downloaded content to. We jettisoned that env var while scrubbing the environment for other secrets that may be lurking.

This PR adds it and a few other important UNIX-y ones back in to Unix environments.


socksy (approved):
I am pretty sure the $PATH one is gonna bite us in the arse soon but let's get this in to fix $HOME

PR #179, opened 2026-01-21
Unify execution interfaces to new one, deprecate old one

socksy:
Standardising on one execution interface. Renames `ExecutionBackend`/`ExecutionHandle` back to `Backend`/`App` (the old names) and adds default method impls so backends don't stub what they don't use. `LocalApp` implements `App` directly now; `SubprocessHandle` sticks around as a thin wrapper because `auto_cleanup`'s temp-dir lifecycle has to hang off something.

Cancellation used to ride in as exit code `-1`, which conflated it with IO errors and signal kills. `execute_local_app` now sends a `Result<Status, Error>` and `Status` grows a `Cancelled` variant. Hopefully this lets the control plane drop the "preserve Cancelled if last_known_status was Cancelled" heuristic.

`Cancelled` is a hard break for exhaustive matches — one-line arm needed in tower-runner when it relinks, tracked out of band. The trait rename is kept deprecated-but-defined (not aliased) so downstream impls compile with a guiding message rather than an unresolved-import error. Root-level `tower_runtime::App` is also deprecated in favour of `execution::App`.

### DIFF
--- a/crates/tower-runtime/src/execution.rs
+++ b/crates/tower-runtime/src/execution.rs
@@ -149,17 +149,19 @@ pub struct NetworkingSpec {
 }
 
 // ============================================================================
-// Execution Backend Trait
+// Backend Trait
 // ============================================================================
 
-/// ExecutionBackend abstracts the compute substrate
+/// Backend creates App instances for a specific compute substrate.
+///
+/// Implementations: SubprocessBackend (subprocess), K8sBackend (Kubernetes)
 #[async_trait]
-pub trait ExecutionBackend: Send + Sync {
-    /// The handle type this backend returns
-    type Handle: ExecutionHandle;
+pub trait Backend: Send + Sync {
+    /// The App type this backend creates
+    type App: App;
 
-    /// Create a new execution environment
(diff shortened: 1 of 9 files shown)


crates/tower-runtime/src/execution.rs
```diff
@@ -195,13 +197,15 @@ pub struct BackendCapabilities {
...
 // ============================================================================
 
-/// ExecutionHandle represents a running execution
+/// App represents a running Tower application instance.
+///
+/// Implementations: LocalApp (subprocess), K8sApp (Kubernetes pod)
 #[async_trait]
-pub trait ExecutionHandle: Send + Sync {
-    /// Get a unique identifier for this execution
+pub trait App: Send + Sync {
```
sammuti: Note this would require a change in the [runner](https://github.com/tower/tower-monorepo/tree/develop/tower-runner) side for the k8s impl. I propose just keeping the new name rather than renaming since I believe it's better name than "App"

bradhe:
I think the main thrust of this PR is to standardize the interfaces slightly better than they are now. I do think that's valid, it just needs to be modernized a bit. I don't have strong opinions about how that should come together however. I'd be open to keeping this around and reintegrating the latest into it.

PR #234, opened 2026-03-25
fix: upgrade http to https in session URL for non-local hosts

socksy:
When logging in with an http:// tower URL against a remote server, subsequent POST requests (like session refresh) get silently downgraded to GET on 301/302 redirects per RFC 7231, causing 404s. This upgrades http to https at login time for non-localhost URLs.


konstantinoscs (approved):
if you tested it and it works!

PR #243, opened 2026-04-07
Use different API key name and fix `tower login --no-browser` test

jo-sm:
This PR updates the expected API key name to be `TOWER_TEST_API_KEY`, since `TOWER_API_KEY` is the actual environment variable we use in the CLI, and fixes an issue with the existing `tower login --no-browser` tests.


bradhe (approved):
I think this is exactly what I was thinking too, nice job. Let's try it out?

PR #244, opened 2026-04-09
feat: add Environment variable column in describe catalog cli output …

konstantinoscs:
Adding a new column in the CLI output of `show catalog` to output the environment variable verbatim as per the API reponse of `describe-catalog`


bradhe (approved):
Looks good to me

PR #248, opened 2026-04-14
feat: multi-environment support for tower deploy

bradhe:
Updates `tower deploy` to support the backend changes for targeting specific environments.

Adds:
- `--environment=<name>` (`-e`) to target a specific environment
- `--all` to deploy to all available environments
- Preserves the previous behavior (deploy to default) when no flag is provided

### DIFF
--- a/crates/tower-cmd/src/deploy.rs
+++ b/crates/tower-cmd/src/deploy.rs
@@ -24,6 +24,20 @@ pub fn deploy_cmd() -> Command {
                 .help("Automatically force creation of the app if it doesn't already exist")
                 .action(clap::ArgAction::SetTrue),
         )
+        .arg(
+            Arg::new("environment")
+                .long("environment")
+                .short('e')
+                .help("The environment to deploy to")
+                .conflicts_with("all"),
+        )
+        .arg(
+            Arg::new("all")
+                .long("all")
+                .help("Deploy to all environments")
+                .action(clap::ArgAction::SetTrue)
+                .conflicts_with("environment"),
+        )
         .about("Deploy your latest code to Tower")
 }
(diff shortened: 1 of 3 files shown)


crates/tower-cmd/src/util/deploy.rs
```diff
@@ -116,7 +119,14 @@ pub async fn deploy_app_package(
 
     // Create the URL for the API endpoint
     let base_url = &api_config.base_path;
-    let url = format!("{}/apps/{}/deploy", base_url, app_name);
+    let url = if all_environments {
+        format!("{}/apps/{}/deploy?all_environments=true", base_url, app_name)
+    } else if let Some(env) = environment {
+        let encoded_environment = urlencode(env);
+        format!("{}/apps/{}/deploy?environment={}", base_url, app_name, encoded_environment)
```
socksy, line 126: could this lead to some environments that are impossible to use via the CLI as we don't url encode them on the other side? (or do we url_decode the env name?)

crates/tower-cmd/src/deploy.rs
```diff
@@ -35,10 +49,30 @@ fn resolve_path(args: &ArgMatches) -> PathBuf {
     }
 }
 
+/// Resolves the target environment from CLI args.
+///
+/// - `--all` → `DeployTarget::All`
+/// - `--environment <name>` → `DeployTarget::Environment(name)`
+/// - neither → `DeployTarget::Default`
```
socksy, line 56: I don't see a `DeployTarget::Default`?

socksy (approved):
Apart from the apparently wrong doc string looks good to me.

PR #257, opened 2026-04-21
refactor(tower-package): extract pure core crate for bundle building

socksy:
- Splits tower-package into a pure core plus native (default, tokio/fs) and wasm (wasm-bindgen) feature shells, so the same deterministic build_package runs from the Rust CLI and from TypeScript.
- Adds tower-package-wasm to the release pipeline via npm OIDC trusted publishing — no long-lived secret to rotate.
- Adds CI/tests


crates/tower-package/Cargo.toml
```diff
@@ -18,4 +18,5 @@ tmpdir = { workspace = true }
 tokio = { workspace = true }
 tokio-stream = { workspace = true }
 tokio-tar = { workspace = true }
+tower-package-core = { workspace = true }
```
bradhe: Why does this have to be in a separate package? We can keep this all in the same package.

bradhe (commented):
Looking good overall, let's actually push hard to get this out--will be a big benefit for folks. Two things to look at:

1. Let's just combine the packages together, and split the implementations. No need to have a separate package, feels to me it just confuses things. We could use feature gates for this.
2. Let's actually go all the way and add this to the release pipeline too, where we push `tower-package-wasm` to NPM. We can then create a wrapper called `tower` that pulls this in for convenient, but at least we can give people the WASM for now.

crates/tower-package/Cargo.toml
socksy: I thought it would be nice to be more flexible later, but not a must for sure - could either use feature flags (the cargo kind) or target_arch conditionals instead

crates/tower-package/test/build.test.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    const body = data.subarray(offset + 512, offset + 512 + size);
+    entries.push({ name, data: body });
+
+    offset += 512 + Math.ceil(size / 512) * 512;
+  }
+  return entries;
+}
+
+function minimalInputs(): BundleInputs {
```
bradhe: Can we drop the name `bundle`? We call it `package` everywhere else--either that, or we need to go update our docs and everything else everywhere 😄 My mind immediately was wondering "what's the difference between a package and a bundle?"

crates/tower-package/test/build.test.ts
```diff
@@ -0,0 +1,140 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { gunzipSync } from "node:zlib";
+
+import {
+  buildBundle,
+  type BundleInputs,
+} from "../pkg/tower_package.js";
```
bradhe, line 8: Another vote in the package direction: `tower_package` exposes a `BundleInputs` type?

crates/tower-package/src/wasm.rs
```diff
@@ -0,0 +1,66 @@
...
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct JsInputs {
+    app_files: Vec<JsEntry>,
+    module_files: Vec<JsEntry>,
+    towerfile_bytes: serde_bytes::ByteBuf,
+    invoke: String,
+    parameters: Vec<Parameter>,
+    schedule: Option<String>,
+    import_paths: Vec<String>,
```
bradhe: All these fields are in the `Towerfile` -- shouldn't we rely on that for them? Also `schedule` should get jettisoned entirely, heh.

bradhe (commented):
Another quick round of feedback. Biggest thing is the `invoke`, `schedule`, etc., are all duplicative of the Towerfile!

bradhe (approved):
Looks good, fix the build failure and let's ship it!

PR #270, opened 2026-04-27
Only require setuptools when an initial attempt fails

bradhe:
We had some overly-aggressive logic that tried to [resolve a known issue with setuptools](https://github.com/pypa/setuptools/issues/5174) in the Python ecosystem. This PR makes that slightly more efficient in the default case with a reasonable fallback when something goes wrong.

### DIFF
--- a/crates/tower-runtime/src/local.rs
+++ b/crates/tower-runtime/src/local.rs
@@ -275,7 +275,49 @@ async fn execute_local_app(
                 ));
 
                 // Let's wait for the setup to finish. We don't care about the results.
-                let res = wait_for_process(ctx.clone(), &cancel_token, child).await;
+                let mut res = wait_for_process(ctx.clone(), &cancel_token, child).await;
+
+                // If the requirements.txt install failed, retry with the legacy
+                // setuptools<82 pin. Some apps (those whose transitive deps rely on
+                // pkg_resources) need that pin to install successfully; we don't
+                // apply it by default because it conflicts with apps whose deps
+                // require setuptools>=82.
+                if res != 0 && uv.should_use_legacy_setuptools_pin(&working_dir) {
+                    let _ = opts.output_sender.send(Output {
+                        channel: Channel::Setup,
+                        fd: FD::Stdout,
+                        line: "tower: dependency install failed; retrying with setuptools<82 pin for pkg_resources compatibility".to_string(),
+                        time: chrono::Utc::now(),
+                    });
+
+                    match uv
+                        .sync_with_legacy_setuptools_pin(&working_dir, &env_vars)
+                        .await
+                    {
+                        Err(e) => {
+                            return Err(e.into());
+                        }
+                        Ok(mut retry_child) => {
+                            let stdout = retry_child.stdout.take().expect("no stdout");
+                            tokio::spawn(drain_output(
+                                FD::Stdout,
+                                Channel::Setup,
+                                opts.output_sender.clone(),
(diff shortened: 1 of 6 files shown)


crates/tower-uv/src/lib.rs
```diff
@@ -357,6 +357,45 @@ impl Uv {
                 &self.uv_path, cwd
             );
 
+            let req_path = cwd.join("requirements.txt");
+
+            // setuptools 82 removed pkg_resources, but many legacy packages still
+            // import it without declaring the dependency. Historically we always
```
sammuti: Why not just remove this `<82` constraint and install `pkg_resources` by default for new versions if that is thing that breaks?

bradhe: I think it's kinda six one way, half dozen the other: Either way, we need to probe to see if it'll work with the default configuration and, if not, change the dependency set. 

I looked in to it directly just now and apparently `pkg_resources` isn't published independently, it's a part of `setuptools`--so it's really a `setuptools` problem, not a `pkg_resources` problem.

crates/tower-uv/src/lib.rs
```diff
@@ -357,6 +357,45 @@ impl Uv {
...
+            // injected `setuptools<82` to keep pkg_resources available, but that
+            // pin makes resolution fail for apps using newer deps that require
+            // setuptools>=82.
+            //
+            // https://github.com/pypa/setuptools/issues/5174
+            let unpinned_resolves = {
+                let mut probe = Command::new(&self.uv_path);
```
sammuti: Actually this solution disguises possibly a major latency increase for the happy path when the user app does not pin setuptools in requirements.txt it seems as there is an additional round trip to pypi to resolve

bradhe: This is typically really fast; however, I generally agree that we have to go to PyPI twice. Do you have another suggestion for how we can do this with only one round trip?

bradhe: I think I have a solution actually. One sec.

sammuti (approved):

crates/tower-runtime/src/local.rs
```diff
@@ -275,7 +275,49 @@ async fn execute_local_app(
...
+
+                    match uv
+                        .sync_with_legacy_setuptools_pin(&working_dir, &env_vars)
+                        .await
+                    {
+                        Err(e) => {
+                            return Err(e.into());
+                        }
+                        Ok(mut retry_child) => {
+                            let stdout = retry_child.stdout.take().expect("no stdout");
```
socksy, line 301: could probably take this whole bit out into its own helper function now, but not a blocker

socksy (approved):

crates/tower-uv/src/lib.rs
```diff
@@ -357,44 +357,85 @@ impl Uv {
...
-            {
-                cmd.process_group(0);
-            }
+    /// Returns whether a failed `sync()` for this directory is eligible for a
+    /// retry via [`sync_with_legacy_setuptools_pin`]. Only applies to projects
+    /// driven by `requirements.txt`; pyproject-based projects manage their own
+    /// setuptools dependency.
+    pub fn should_use_legacy_setuptools_pin(&self, cwd: &Path) -> bool {
+        cwd.join("requirements.txt").exists()
+    }
```
socksy, line 373: I think if we're doing the retesting method we have no reason to differentiate

PR #272, opened 2026-04-28
fix: extend setuptools<82 retry to pyproject.toml apps

socksy:
## Summary

The legacy-setuptools retry added in #270 only kicked in for `requirements.txt` projects — pyproject.toml apps that needed `pkg_resources` (setuptools<82) at install time had no recovery path and just failed.

`uv sync` can't accept a CLI constraint, so the retry now drops out of `uv sync` into `uv pip install` with the project source (`.` for pyproject, `-r requirements.txt` otherwise) plus a `setuptools<82` arg. Both kinds of project get the same fallback.

The call site in tower-runtime is now `if res != 0 { retry }` — no file-type knowledge — and the file-type branch lives inline in `sync_with_legacy_setuptools_pin` where the command args are built. `should_use_legacy_setuptools_pin` and `spawn_requirements_install` are gone; a small `run_setup_child` helper dedupes the stdout/stderr drain + wait between the initial install and the retry.

## Test plan

- [x] `cargo test -p tower-uv --test sync_test` (7 tests, including two new pyproject ones)
- [x] Verified the new pyproject tests fail when production code is reverted (negative control)

### DIFF
--- a/crates/tower-runtime/src/local.rs
+++ b/crates/tower-runtime/src/local.rs
@@ -256,33 +256,15 @@ async fn execute_local_app(
                     }
                 }
             }
-            Ok(mut child) => {
-                // Drain the logs to the output channel.
-                let stdout = child.stdout.take().expect("no stdout");
-                tokio::spawn(drain_output(
-                    FD::Stdout,
-                    Channel::Setup,
-                    opts.output_sender.clone(),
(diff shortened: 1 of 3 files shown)


sammuti (approved):
LGTM

PR #278, opened 2026-05-13
Align api cli, and mcp

codingcyclist:
Context:

The API scopes apps, catalogs and secrets to environments, and the CLI `tower deploy` command already supports --environment / -e (defaulting to "default"). However:
 1. CLI tower apps list — has no --environment flag; always passes environment: None to the API
 2. MCP tower_deploy — hardcodes "default" environment with no way for the caller to specify one
 3. MCP tower_apps_list — has no environment parameter at all
 4. MCP tower_run_remote — hardcodes let env = "default" with no way to override
 5. MCP has not catalog list / show support

Specifically, the shortcomings on the MCP don't allow for critical use cases (deploy apps to specific environments, run apps in specific environments, etc.)

Note: The secrets MCP tools (tower_secrets_list, tower_secrets_create, tower_secrets_delete) already correctly handle environments. There are no
 catalog MCP tools currently, and the CLI catalogs list already supports --environment.

The desired behavior:
1. all list/show/run commands support an optional --environment flag
2. all list/show/run commands default to None (i.e. all environments) environment unless overridden.
3. MCP exposes same catalog discovery tools as CLI

### DIFF
--- a/crates/tower-cmd/src/api.rs
+++ b/crates/tower-cmd/src/api.rs
@@ -116,11 +116,11 @@ pub async fn describe_app(
 
 pub async fn list_apps(
     config: &Config,
-    environment: &str,
+    environment: Option<&str>,
 ) -> Result<Vec<tower_api::models::AppSummary>, Error<tower_api::apis::default_api::ListAppsError>>
 {
     let api_config: configuration::Configuration = config.into();
-    let environment = environment.to_string();
+    let environment = environment.map(|s| s.to_string());
 
     fetch_all_pages(|page, page_size| {
         let api_config = &api_config;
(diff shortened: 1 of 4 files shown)


crates/tower-cmd/src/api.rs
```diff
@@ -115,12 +116,15 @@ pub async fn describe_app(
 
 pub async fn list_apps(
     config: &Config,
+    environment: Option<&str>,
 ) -> Result<Vec<tower_api::models::AppSummary>, Error<tower_api::apis::default_api::ListAppsError>>
 {
     let api_config: configuration::Configuration = config.into();
+    let environment = environment.map(|s| s.to_string());
```
bradhe, line 123: nit: This could probably be promoted to a function call and clean up this diff a bit.

bradhe (approved):
Generally looks good to me!

PR #281, opened 2026-05-17
Fix local PYTHONPATH join

bradhe:
Running `tower run --local` or `tower run` in a remote environment against an app whose Towerfile declares two or more `import_paths` panicked the runtime:

```
thread 'tokio-runtime-worker' panicked at crates/tower-runtime/src/local.rs:
called `Result::unwrap()` on an `Err` value: JoinPathsError
Oh no! Your local run crashed with exit code: -1
```

The issue is that in `crates/tower-runtime/src/local.rs`, `PYTHONPATH` was built in two stages:

1. The Towerfile's `import_paths` were joined via `std::env::join_paths` into a single colon-separated string and stored in the env map under `PYTHONPATH`.
2. Later, `make_env_vars` appended the working directory by calling `std::env::join_paths([existing_pythonpath, cwd]).unwrap()`.

`join_paths` rejects any component that already contains the platform path separator, so step 2 failed with `JoinPathsError` and the `unwrap()` panicked. The same shape of bug existed at the earlier append site for user-supplied `PYTHONPATH` env vars.

## The fix

import_paths is now threaded through to make_env_vars as a &[PathBuf], and make_env_vars composes the final PYTHONPATH in a single env::join_paths call over (existing entries split with split_paths) + import_paths + cwd. No more round-tripping a joined string back through join_paths, and no intermediate helper.

As a defense-in-depth measure, Towerfile::from_toml now also rejects import_paths entries that are empty or contain :/;, surfacing an InvalidTowerfile error instead of letting bad input fail deep inside the runtime.

A regression test plus a new 06-multiple-import-paths example app exercise the multi-import-path path, and the original repro now runs cleanly.

### DIFF
--- a/crates/tower-runtime/src/local.rs
+++ b/crates/tower-runtime/src/local.rs
@@ -161,31 +161,17 @@ async fn inner_execute_local_app(
     let manifest = &package.manifest;
     let secrets = opts.secrets;
     let params = opts.parameters;
-    let mut other_env_vars = opts.env_vars;
+    let other_env_vars = opts.env_vars;
 
-    if !package.manifest.import_paths.is_empty() {
-        debug!(ctx: &ctx, "adding import paths to PYTHONPATH: {:?}", package.manifest.import_paths);
+    let import_paths: Vec<PathBuf> = package
+        .manifest
+        .import_paths
+        .iter()
+        .map(|p| package_path.join(p))
+        .collect();
 
-        let import_paths = package
-            .manifest
-            .import_paths
-            .iter()
-            .map(|p| package_path.join(p))
-            .collect::<Vec<_>>();
-
-        let import_paths = std::env::join_paths(import_paths)?
-            .to_string_lossy()
-            .to_string();
-
-        if let Some(existing) = other_env_vars.get("PYTHONPATH") {
-            // A user-supplied PYTHONPATH may itself contain multiple entries, so we split it
-            // before re-joining to avoid `join_paths` rejecting the pre-joined string.
-            let pythonpath = append_to_path_list(existing, &import_paths);
-            other_env_vars.insert("PYTHONPATH".to_string(), pythonpath);
-        } else {
-            // Otherwise, we just set it.
-            other_env_vars.insert("PYTHONPATH".to_string(), import_paths);
-        }
+    if !import_paths.is_empty() {
+        debug!(ctx: &ctx, "adding import paths to PYTHONPATH: {:?}", import_paths);
     }
 
     // We insert these checks for cancellation along the way to see if the process was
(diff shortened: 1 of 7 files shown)


crates/tower-runtime/src/local.rs
```diff
@@ -177,13 +177,10 @@ async fn inner_execute_local_app(
...
-            // If we already have a PYTHONPATH, we need to append to it.
-            let existing = other_env_vars.get("PYTHONPATH").unwrap();
-            let pythonpath = std::env::join_paths(vec![existing, &import_paths])?
-                .to_string_lossy()
-                .to_string();
-
+        if let Some(existing) = other_env_vars.get("PYTHONPATH") {
+            // A user-supplied PYTHONPATH may itself contain multiple entries, so we split it
+            // before re-joining to avoid `join_paths` rejecting the pre-joined string.
+            let pythonpath = append_to_path_list(existing, &import_paths);
```
socksy: Rather than serializing to a string, then immediately splitting it and then reserializing, wouldn't it make more sense to make the PYTHONPATH in one join pass in `make_env_vars` (i.e. the call site)?

socksy (changes requested):
AFAICT, this is a lot of machinery that doesn't actually work atm 😅

1. We create a string for the `PYTHONPATH` separately from all the other env vars
2. We immediately split the string back into a vector
3. We conditionally check to see if `other_env_vars` has `PYTHONPATH`, and if so append the `cwd` to it

However
- I don't think it makes sense to serialize to a string and then split it again, why not just add it to the vec and then serialize?
- `other_env_vars` can never have `PYTHONPATH` set, since we never set it before this code is called (afaict?)
- the user's `PYTHONPATH` env var never makes it down the call stack to here — and I think it's maybe arguable whether it should
    - In the case of a local CLI run, it makes sense to keep it clean, so people don't think they'll have a library they never set up
    - OTOH for self hosted runners we may want users to be able to pass in custom `PYTHONPATH`s to have predownloaded deps. I suspect this should be a different env var, and something we worry about if someone requests this feature

My suspicion that the solution for this bug is to rather delete all this PYTHONPATH special casing and just use a single `join_paths` with `PWD` + `import_paths` in `make_env_vars`

crates/tower-runtime/src/local.rs
bradhe: Good call — refactored in 1a15375. `make_env_vars` now takes `import_paths: &[PathBuf]` and composes PYTHONPATH (existing entries split, import_paths, cwd) in one `env::join_paths`. `append_to_path_list` is gone.

bradhe:
@socksy it fixes the bug--the regression tests verifies that, and running it locally it's very easy to reproduce; however, I think your optimizations are good to include. I've updated the PR to address your feedback.

socksy (approved):

PR #285, opened 2026-06-01
fix(tower-runtime): preserve error details in SpawnFailed and PackageUnpackFailed

sammuti:
## Summary
- Add `detail: String` field to `Error::SpawnFailed` and `Error::PackageUnpackFailed` so the root cause (IO error, missing uv, permission denied, etc.) flows through to termination logs and trace spans
- Add `Display` impl for `tower_uv::Error`
- Previously all `tower_uv::Error` variants were mapped to a bare `Error::SpawnFailed` with no context, making production `spawn_failed` errors impossible to diagnose without pod-level access

## Context

Investigating TOW-2151/TOW-2160 (WaiterClosed) and recent `spawn_failed` errors on May 31, we found that while the structured `AppCompletion::Failed` pipeline (from #274) correctly surfaces the error kind, the actual error *message* was always just "SpawnFailed" because the `From<tower_uv::Error>` impl discarded the inner detail.

After this change, termination logs and trace spans will show messages like:
- `spawning process: not found: uv binary not found in PATH`
- `spawning process: IO error: No such file or directory (os error 2)`
- `spawning process: permission denied: /app/.venv/bin/python`
- `package unpack failed: InvalidManifest`

## Test plan
- [x] `cargo check --workspace` passes
- [x] `cargo test --workspace` passes
- [ ] Deploy to staging and trigger a run with a broken pyproject.toml to verify the detail appears in traces

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages with detailed context for spawn failures and package operations, helping users better understand what went wrong during runtime execution.

### DIFF
--- a/crates/tower-runtime/src/errors.rs
+++ b/crates/tower-runtime/src/errors.rs
@@ -5,8 +5,8 @@ pub enum Error {
     #[snafu(display("failed to RPC server"))]
     RuntimeStartFailed,
 
-    #[snafu(display("spawning process"))]
-    SpawnFailed,
+    #[snafu(display("spawning process: {detail}"))]
+    SpawnFailed { detail: String },
 
     #[snafu(display("no app running"))]
     NoRunningApp,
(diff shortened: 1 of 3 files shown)


bradhe (approved):
Looks good. Just fix failing CI.

PR #286, opened 2026-06-02
fix: keep --json output parseable on stdout

socksy:
`--json` is supposed to mean "stdout is JSON you can parse". Several commands broke that promise by mixing raw text into the output. `tower run --detached --json` printed a "See more" link after the success JSON. `tower run --json` against a missing app printed remediation steps after the error JSON. `tower schedules list --json` printed "No schedules found." instead of `[]`. `tower teams list --json` printed the active-team legend after the data. And `tower version --json` wasn't JSON at all.

The root cause is the same everywhere: `output::write()` knows about MCP mode but not JSON mode. Instead of special-casing call sites, the broken sites now go through JSON-aware output functions. output.rs gains two: `text()` writes a human rendering or the data as JSON (the same shape as `table()` and `list()`), and `note()` writes presentation-only text that JSON mode suppresses. The run link travels inside the success message, with the scheduled run as structured `data`.

The new integration tests run each case with `--json` and parse stdout. Putting `teams list` under test required updating the mock server's `/v1/session/refresh` response, which had drifted from the generated client models. flake.lock gets a nixpkgs bump so binaryen comes from the binary cache on aarch64-darwin instead of compiling LLVM from source.

### DIFF
--- a/crates/tower-cmd/src/output.rs
+++ b/crates/tower-cmd/src/output.rs
@@ -487,6 +487,25 @@ pub fn list<T: Serialize>(items: Vec<String>, json_data: Option<&T>) {
     }
 }
 
+/// Writes a human-readable rendering of some data, or the data itself as JSON when
+/// in JSON mode. Use this when a command's output is data that has both a plain text
+/// and a JSON representation, mirroring `table` and `list`.
+pub fn text<T: Serialize>(msg: &str, json_data: &T) {
+    if get_output_mode().is_json() {
+        json(json_data);
+    } else {
+        write(msg);
+    }
+}
+
+/// Writes presentation-only text that accompanies human-formatted output, like table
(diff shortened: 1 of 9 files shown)


crates/tower-cmd/src/output.rs
```diff
@@ -487,6 +487,25 @@ pub fn list<T: Serialize>(items: Vec<String>, json_data: Option<&T>) {
     }
 }
 
+/// Writes a human-readable rendering of some data, or the data itself as JSON when
+/// in JSON mode. Use this when a command's output is data that has both a plain text
+/// and a JSON representation, mirroring `table` and `list`.
+pub fn text<T: Serialize>(msg: &str, json_data: &T) {
```
konstantinoscs, line 493: Ah ok I made something similar in my PR and after this I can probably remove it

konstantinoscs (approved):
Great!

PR #288, opened 2026-06-08
feat: publish Tower as an installable Claude Code plugin

socksy:
The repo already had a Claude Code plugin (plugin.json + skills/tower/SKILL.md + .mcp.json) but no way for anyone to install it. A plugin only becomes installable once a marketplace catalog lists it, so this adds one.

## Changes

- Add `.claude-plugin/marketplace.json` at the repo root, cataloguing the `tower` plugin with `source: "./plugin"`
- Move the plugin (`plugin.json`, `skills/`, `.mcp.json`) into a `plugin/` subdirectory so installing it caches only the plugin, not the entire tower-cli repo
- Point the `regenerate-skill` workflow at the new `plugin/skills/tower/SKILL.md` path

`plugin.json` is unchanged: it relies on the default discovery of `skills/` and `.mcp.json` relative to the plugin root, which still resolve after the move.

## Install flow (once merged to the default branch)

```
/plugin marketplace add tower/tower-cli
/plugin install tower@tower-cli
```

## Note

The root `.mcp.json` moved into `plugin/`. It was not gitignored, so it also auto-loaded the `tower` MCP server for anyone doing dev work in tower-cli itself. If that project-scoped behaviour is wanted, we can restore a copy at the root in a follow-up.


plugin/.mcp.json
giray123, line 1: (optional)There is a line at SKILL.md : ""copy the .mcp.json from the tower-cli repo into your project root...", may be should be now `plugin/.mcp.json`?

.github/workflows/regenerate-skill.yml
giray123, line 1: (optional)
> `tower skill generate` output has drifted from the checked-in `skills/tower/SKILL.md`

`plugin/skills/...` ?

giray123 (approved):

PR #293, opened 2026-06-11
Add environments delete command

jo-sm:
This PR adds a new `environments delete` command to the CLI. Depends on #292.

### DIFF
--- a/crates/tower-cmd/src/api.rs
+++ b/crates/tower-cmd/src/api.rs
@@ -1021,6 +1021,17 @@ impl ResponseEntity for tower_api::apis::default_api::ListEnvironmentsSuccess {
     }
 }
 
+impl ResponseEntity for tower_api::apis::default_api::DeleteEnvironmentSuccess {
+    type Data = tower_api::models::DeleteEnvironmentResponse;
+
+    fn extract_data(self) -> Option<Self::Data> {
+        match self {
+            Self::Status200(resp) => Some(resp),
+            Self::UnknownValue(_) => None,
+        }
+    }
+}
+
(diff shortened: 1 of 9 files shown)


crates/tower-cmd/src/api.rs
```diff
@@ -1021,6 +1021,17 @@ impl ResponseEntity for tower_api::apis::default_api::ListEnvironmentsSuccess {
     }
 }
 
+impl ResponseEntity for tower_api::apis::default_api::DeleteEnvironmentSuccess {
```
socksy, line 1035: this file is autogenerated, right? If not, I hate it :D

crates/tower-cmd/src/environments.rs
```diff
@@ -50,3 +50,18 @@ pub async fn do_create(config: Config, args: &ArgMatches) {
...
 }
+
+pub async fn do_delete(config: Config, args: &ArgMatches) {
+    let name = args.get_one::<String>("name").unwrap_or_else(|| {
+        output::die("Environment name (--name) is required");
+    });
+
+    output::with_spinner(
+        "Deleting environment",
+        api::delete_environment(&config, name),
```
socksy: IMO we should definitely have a confirmation here, especially because idiot LLMs love to drive CLI tools

socksy (commented):
Deleting itself seems fine, but I'm pretty sure we should have an interactive prompt and/or `--dangerously-confirm-force-deletion-of-environment` (or you know, something snappier) before I would approve

konstantinoscs (approved):
LGTM!

PR #296, opened 2026-06-15
fix: c tls be gone for good

konstantinoscs:
Switch `tower-cli`'s rustls crypto provider from `aws-lc-rs` to `ring`.

The runner already uses rustls/tonic with the `ring` provider, but `tower-cli` was still enabling `rustls`'s `aws-lc-rs` feature. Because `tower-runner` depends on crates from `tower-cli`, that pulled `aws-lc-rs` and `aws-lc-sys` into the runner dependency graph.

`aws-lc-sys` builds AWS-LC C code and trips over the GCC memcmp compiler-bug check in the Linux cross build images (see what https://github.com/tower/tower-monorepo/pull/1729 tries to solve). Moving `tower-cli` onto the `ring` provider removes `aws-lc-rs`/`aws-lc-sys` from the graph and avoids the C-toolchain cross-compilation shitstorm entirely.

## Summary by CodeRabbit

* **Chores**
  * Updated cryptographic provider backend configuration for improved compatibility and performance across the application.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -201,28 +201,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
-name = "aws-lc-rs"
-version = "1.16.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


konstantinoscs:
@coderabbitai I triggered a review you dumb. Cancel the other PR and review this one

jo-sm (approved):

PR #298, opened 2026-06-16
[TOW-2269] Validate invoke script exists in from_unpacked_path function

sammuti:
## Release Notes

* **Bug Fixes**
  * Enhanced package initialization to validate required scripts and provide clearer error messages when script files cannot be found in the project.

### DIFF
--- a/crates/tower-package/src/native.rs
+++ b/crates/tower-package/src/native.rs
@@ -100,6 +100,21 @@ impl Package {
         file.read_to_string(&mut contents).await?;
         let manifest = Manifest::from_json(&contents)?;
 
+        // Validate the invoke script exists on disk.
+        if !manifest.invoke.is_empty() {
+            let working_dir = if manifest.version == Some(1) {
+                path.clone()
+            } else {
+                path.join(&manifest.app_dir_name)
+            };
+            let invoke_path = working_dir.join(&manifest.invoke);
+            if !invoke_path.exists() {
+                return Err(Error::MissingScript {
+                    script: manifest.invoke,
+                });
+            }
+        }
+
         Ok(Self {
             tmp_dir: None,
             package_file_path: None,
--- a/crates/tower-package/src/native.rs
+++ b/crates/tower-package/src/native.rs
@@ -102,6 +102,15 @@ impl Package {
 
         // Validate the invoke script exists on disk.
         if !manifest.invoke.is_empty() {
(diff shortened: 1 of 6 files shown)


crates/tower-package/src/native.rs
```diff
@@ -100,6 +100,21 @@ impl Package {
...
+            } else {
+                path.join(&manifest.app_dir_name)
+            };
+            let invoke_path = working_dir.join(&manifest.invoke);
+            if !invoke_path.exists() {
+                return Err(Error::MissingScript {
+                    script: manifest.invoke,
+                });
+            }
+        }
```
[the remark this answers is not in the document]
sammuti: check now

crates/tower-package/src/native.rs
```diff
@@ -100,6 +100,8 @@ impl Package {
         file.read_to_string(&mut contents).await?;
         let manifest = Manifest::from_json(&contents)?;
 
+        Self::validate_invoke_script(&manifest, &path)?;
```
konstantinoscs, line 103: `tower deploy` builds with Package::build in deploy.rs (line 128) but the new validation only runs here. I manually verified tower package succeeds with script = "main.py" and source = ["*.txt"]; the tar contained only app/readme.txt, MANIFEST, and Towerfile so the runner-side UX improves, but the CLI still uploads invalid bundles.

konstantinoscs (commented):
I have a comment and then I also ran an agent which produced the following two points as well:

1. The path validation is too blunt and incomplete.
native.rs-l:121 rejects any string containing "..", which blocks valid filenames like model..v2.py, and native.rs-l:139 uses exists(), so a directory can pass as the script. Prefer component-based validation for ParentDir / absolute roots and require metadata.is_file().

2. Shell script path validation does not match shell runtime execution.
The validator checks working_dir.join(invoke) at native.rs-l:138, and Python runtime also uses working_dir.join(invoke) at local.rs-l:221. But shell runtime invokes package_path.join(invoke) at local.rs-l:490. For .sh apps, the validator can accept a path the runtime will not execute, or reject existing workarounds that include app/.

Worth running thourgh your agent as well

@coderabbitai wdyt?

crates/tower-package/src/native.rs
sammuti: Good point.

bradhe (approved):
Looks good, let's get it out.

PR #303, opened 2026-06-18
feat: send X-Tower-Idempotency-Key on deploy

bradhe:
## Summary

Wires the `tower deploy` command to send an `X-Tower-Idempotency-Key` header so consecutive deploys of identical source (e.g. to `staging` then `production`) reuse a single `AppVersion` server-side instead of creating a new version each time. The server already honors this header; the CLI just didn't send it.

## Behavior

1. **Auto-populate from git when safe.** If the deploy directory is inside a git worktree:
   - Clean tree → send `X-Tower-Idempotency-Key: <git HEAD SHA>`.
   - Dirty tree → omit the header, so every deploy creates a new version (provenance is never misrepresented).
2. **Manual override.** `--idempotency-key <value>` wins over auto-detection — useful for CI that builds artifacts outside a git checkout.
3. **Opt-out.** `--no-idempotency-key` suppresses the header even on a clean tree (conflicts with `--idempotency-key`).
4. **Stdout hint.** When the server returns a version whose key matches what we sent and whose `created_at` predates this deploy, the CLI prints a line explaining that an existing version was reused (suppressed in JSON mode).

The MCP `tower_deploy` tool gets the same git auto-detection.

## Implementation notes

- New `util::git::clean_head_sha()` shells out to `git` (no new dependencies); returns `None` when not in a repo, the tree is dirty, or `git` is unavailable.
- Key flows `deploy_from_dir → do_deploy_package → deploy_app_package → upload_file_with_progress`, which attaches the header when present.


crates/tower-cmd/src/deploy.rs
```diff
@@ -172,22 +218,72 @@ async fn do_deploy_package(
...
+    let created_at = DateTime::parse_from_rfc3339(&version.created_at)
+        .ok()?
+        .with_timezone(&Utc);
+
+    if Utc::now().signed_duration_since(created_at) < REUSE_AGE_THRESHOLD {
+        return None;
+    }
+
+    Some(format!(
+        "Reusing version `{}` (deployed {}) — no source changes since key {}.",
```
socksy: since this is not actually true, but is rather that the git repository doesn't have any committed changes, we should probably mention "git"

crates/tower-cmd/src/deploy.rs
```diff
@@ -172,22 +218,72 @@ async fn do_deploy_package(
                 ),
             };
             output::success(&line);
+
+            // Only emit the informational hint in normal mode; in JSON mode a
+            // bare line would corrupt the structured output.
+            if output::get_output_mode().is_normal() {
```
socksy: FWIW I think if you're using this function to work out what to write out, then it's almost certainly a code smell (it's public so that we can check for `ctrl-c` behaviour differently, but wanting to only print things out if it's not json or mcp mode seems to me like an `output.rs` specific concern). 

Additionally, in this case would it make sense to either re-use `output::success`, or to print it to stderr with `output::write_to_stderr`? If not, perhaps a simple function addition to `output` that writes only informational text to `output_mode().is_normal()` texts would be better?

crates/tower-cmd/src/mcp.rs
```diff
@@ -812,7 +812,20 @@ impl TowerService {
         let env = request.environment.unwrap_or_else(|| "default".to_string());
         let deploy_target = deploy::DeployTarget::Environment(env);
 
-        match deploy::deploy_from_dir(self.config.clone(), working_dir, true, deploy_target).await {
+        // Auto-detect the idempotency key from git (clean tree HEAD) just like
+        // the CLI deploy command does, so repeated deploys of unchanged source
+        // collapse to a single AppVersion server-side.
+        let idempotency_key = crate::util::git::clean_head_sha(&working_dir);
```
socksy, line 818: would be good to do a SHA of the SHAs for the idempotency_key when git is dirty. Alternatively, swap the warning around to tell the user that because their git repo is not clean, the whole bundle is being re-uploaded (current way around you'd only know this behaviour exists if you pushed a clean repo, and given we state that even an untracked file is enough to make a repo dirty, some people may never hit upon this behaviour and wonder why tower doesn't support it)

socksy (approved):
Please have a look at the `output` function comment, but otherwise looks good.

crates/tower-cmd/src/mcp.rs
bradhe: I think we should err on the side of folks properly using git for the time being. The future state will be this is _always_ up to date with Git anyway. I'm happy to say "you called it" if we get complaints about this behavior, which is strictly more limiting than less.

PR #307, opened 2026-06-25
regen clients

konstantinoscs:
what the title says


konstantinoscs:
merging this without an approval as it's ridiculously trivial

PR #309, opened 2026-06-29
feat(cli+sdk): Tower Storage from the CLI — credential vending & connection snippets

konstantinoscs:
## Summary

Adds Tower-vended catalog credentials to the CLI and Python SDK so users can access Tower-managed Iceberg catalogs without manually configuring PyIceberg credentials.

This introduces a vend-first path for `tower.tables()`, adds `tower catalogs credentials`, and improves catalog discovery with storage-focused filters.

## Changes

- Added Python SDK support for Tower-vended Iceberg catalog credentials.
- Updated `tower.tables()` so string catalog names use Tower credential vending by default.
- Added lazy write escalation: reads vend `read` credentials first, while writes re-vend `read-write` credentials and reload the table before writing.
- Kept custom PyIceberg `Catalog` objects working for tests and advanced/custom usage.
- Added a temporary `tower_credentials=False` rollback path for legacy PyIceberg config loading.
- Added credential caching with refresh before token expiry.
- Added default-catalog retry handling for first-time provisioning races.
- Updated Tower context handling to prefer runtime environment values and support logged-in session auth from the local Tower config.
- Added `tower catalogs list --type <type>` and `tower catalogs list --storage`.
- Added `tower catalogs credentials <name>` with snippets for:
  - PyIceberg
  - Spark
  - DuckDB
  - dbt
  - DBeaver
- Made normal credential output avoid printing tokens by default, using `TOWER_CATALOG_TOKEN` instead.
- Added `--show-token` for explicit token printing and `--json` for raw machine-readable output.

New commands and command variations:
```
# --- Catalog discovery (new flags on `catalogs list`) ---
tower catalogs list --storage                       # only Tower-managed (tower-catalog) catalogs
tower catalogs list --type tower-catalog            # same, explicit
tower catalogs list --type s3-tables -e production  # filter by any type, scoped to an environment

# --- Vend short-lived credentials + print connection snippets (new subcommand) ---
tower catalogs credentials <name>                          # all snippets, read-only, token redacted
tower catalogs credentials <name> --format pyiceberg       # one tool: pyiceberg | spark | duckdb | dbt | dbeaver | all
tower catalogs credentials <name> --mode read-write        # read-write creds (requires catalog write permission)
tower catalogs credentials <name> -e production            # choose the environment (default: "default")
tower catalogs credentials <name> --show-token             # inline the real token instead of $TOWER_CATALOG_TOKEN
tower --json catalogs credentials <name>                   # raw JSON (includes the token) for scripting
```

New sdk behavior:
```
import tower

# String catalog name → Tower-vended credentials (laptop, CI, or in a run)
df = tower.tables("events").load().read()              # read-only creds
tower.tables("events").insert(data)                    # escalates to read-write on write

tower.tables("events", tower_credentials=False)        # opt out → legacy local PyIceberg config
tower.tables("events", catalog=my_pyiceberg_catalog)   # custom Catalog object, unchanged
```

Some local testing:
List storage catalogs only (catalog type == tower-catalog)
image

Show a catalog:
image

Vend credentials for a catalog:
```
cargo run -- catalogs credentials default1 --environment default --mode read
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.35s
     Running `target/debug/tower catalogs credentials default1 --environment default --mode read`
✔ Vending catalog credentials... Done!
Catalog: default1
Mode: read
Expires: 2026-06-29T17:24:04.613338+02:00
Token: not printed; snippets read $TOWER_CATALOG_TOKEN
  These credentials are short-lived and intended for ad-hoc development use.

Shell setup
export TOWER_CATALOG_TOKEN="$(tower --json catalogs credentials 'default1' --environment 'default' --mode 'read' | python3 -c 'import json,sys; print(json.load(sys.stdin)["credentials"]["oauth_token"])')"

PyIceberg
import os
from pyiceberg.catalog import load_catalog

catalog = load_catalog(
    "default1",
    type="rest",
    uri="<tower's_very_secret_polaris_instance>",
    warehouse="<fully_qualified_catalog_name>",
    token=os.environ["TOWER_CATALOG_TOKEN"],
)


Spark
import os
from pyspark.sql import SparkSession

spark = (
    SparkSession.builder
    .config("spark.sql.catalog.default1", "org.apache.iceberg.spark.SparkCatalog")
    .config("spark.sql.catalog.default1.type", "rest")
    .config("spark.sql.catalog.default1.uri", "<tower's_very_secret_polaris_instance>")
    .config("spark.sql.catalog.default1.warehouse", "<fully_qualified_catalog_name>")
    .config("spark.sql.catalog.default1.token", os.environ["TOWER_CATALOG_TOKEN"])
    .getOrCreate()
)


DuckDB
duckdb <<SQL
INSTALL httpfs;
LOAD httpfs;
INSTALL iceberg;
LOAD iceberg;
SET s3_region='eu-central-1';
CREATE OR REPLACE SECRET tower_cat (TYPE iceberg, TOKEN '${TOWER_CATALOG_TOKEN}');
ATTACH '<fully_qualified_catalog_name>' AS "default1" (TYPE iceberg, SECRET tower_cat, ENDPOINT '<tower's_very_secret_polaris_instance>', DEFAULT_REGION 'eu-central-1');
SQL


dbt
type: iceberg
catalog_type: rest
uri: "<tower's_very_secrect_polaris_instance>"
warehouse: "<fully_qualified_catalog_name>"
token: "{{ env_var('TOWER_CATALOG_TOKEN') }}"


DBeaver
Catalog type: Iceberg REST
URI: <tower's_very_secret_polaris_instance>
Warehouse: <fully_qualified_catalog_name>
Authentication: Bearer token
Token: $TOWER_CATALOG_TOKEN
```

## Summary by CodeRabbit

* **New Features**
  * Added a new `catalogs credentials` command to generate short-lived catalog credentials and ready-to-use connection snippets.
  * Added catalog filtering to `catalogs list`, including a new `--type` option and a `--storage` shortcut.
  * Expanded supported snippet formats for tools like DuckDB, dbt, PyIceberg, and Spark.

* **Bug Fixes**
  * Improved catalog and table access so write operations automatically switch to the needed access mode.
  * Enhanced session-based settings lookup for more reliable CLI behavior when environment variables are missing.

### DIFF
--- a/crates/tower-cmd/src/catalogs.rs
+++ b/crates/tower-cmd/src/catalogs.rs
@@ -1,7 +1,9 @@
 use clap::{value_parser, Arg, ArgAction, ArgMatches, Command};
 use colored::Colorize;
 use config::Config;
-use tower_api::models::{vend_catalog_credentials_body, CatalogCredentials};
+use tower_api::models::{
+    vend_catalog_credentials_body, CatalogCredentials, DescribeCatalogResponse,
+};
 
 use crate::{api, output, util::cmd};
 
@@ -162,17 +164,16 @@ pub async fn do_credentials(config: Config, args: &ArgMatches) {
     )
     .await;
 
-    output::render_human_or_json(&response, || {
-        print_credentials(
-            name,
-            &env,
-            mode,
-            config.tower_url.as_str(),
-            &response.credentials,
-            format,
-            show_token,
-        );
-    });
+    let human = credentials_text(
+        name,
+        &env,
+        mode,
+        config.tower_url.as_str(),
+        &response.credentials,
+        format,
+        show_token,
+    );
+    output::text(&human, &response);
 }
 
 pub async fn do_show(config: Config, args: &ArgMatches) {
@@ -183,35 +184,8 @@ pub async fn do_show(config: Config, args: &ArgMatches) {
 
     match api::describe_catalog(&config, name, &env).await {
         Ok(response) => {
-            output::render_human_or_json(&response, || {
-                let catalog = &response.catalog;
-
-                output::detail("Catalog", &catalog.name);
-                output::detail("Type", &catalog.r#type);
-                output::detail("Environment", &catalog.environment);
-
-                if !catalog.properties.is_empty() {
-                    output::newline();
-                    output::header("Properties");
-
-                    let headers = vec!["Name", "Runtime Var", "Preview"]
-                        .into_iter()
-                        .map(str::to_string)
-                        .collect();
-                    let data = catalog
-                        .properties
-                        .iter()
-                        .map(|prop| {
-                            vec![
-                                prop.name.clone(),
-                                prop.environment_variable.clone().unwrap_or_default(),
-                                prop.preview.dimmed().to_string(),
-                            ]
-                        })
-                        .collect();
-                    output::table(headers, data, Some(&response));
-                }
-            });
+            let human = catalog_details_text(&response);
+            output::text(&human, &response);
         }
         Err(err) => output::tower_error_and_die(err, "Fetching catalog details failed"),
     }
@@ -250,7 +224,48 @@ fn token_export_command(name: &str, environment: &str, mode: &str, tower_url: &s
     )
 }
 
-fn print_credentials(
+fn detail_line(label: &str, value: &str) -> String {
+    format!("{} {}\n", format!("{}:", label).bold().green(), value)
+}
+
+fn header_line(text: &str) -> String {
+    format!("{}\n", text.bold().green())
+}
+
+fn catalog_details_text(response: &DescribeCatalogResponse) -> String {
+    let catalog = &response.catalog;
+    let mut out = String::new();
+
+    out.push_str(&detail_line("Catalog", &catalog.name));
+    out.push_str(&detail_line("Type", &catalog.r#type));
+    out.push_str(&detail_line("Environment", &catalog.environment));
+
+    if !catalog.properties.is_empty() {
+        out.push('\n');
+        out.push_str(&header_line("Properties"));
+
+        let headers = vec!["Name", "Runtime Var", "Preview"]
+            .into_iter()
+            .map(str::to_string)
+            .collect();
+        let data = catalog
+            .properties
+            .iter()
+            .map(|prop| {
+                vec![
+                    prop.name.clone(),
+                    prop.environment_variable.clone().unwrap_or_default(),
+                    prop.preview.dimmed().to_string(),
+                ]
+            })
+            .collect();
+        out.push_str(&output::table_text(headers, data));
+    }
+
+    out
+}
+
+fn credentials_text(
     name: &str,
     environment: &str,
     mode: &str,
@@ -258,32 +273,42 @@ fn print_credentials(
     credentials: &CatalogCredentials,
     format: &str,
     show_token: bool,
-) {
-    output::detail("Catalog", name);
-    output::detail("Mode", &credentials.mode);
-    output::detail("Expires", &credentials.expires_at);
+) -> String {
+    let mut out = String::new();
+
+    out.push_str(&detail_line("Catalog", name));
+    out.push_str(&detail_line("Mode", &credentials.mode));
+    out.push_str(&detail_line("Expires", &credentials.expires_at));
     if show_token {
-        output::detail("Token", &credentials.oauth_token);
+        out.push_str(&detail_line("Token", &credentials.oauth_token));
     } else {
-        output::detail("Token", "not printed; snippets read $TOWER_CATALOG_TOKEN");
+        out.push_str(&detail_line(
+            "Token",
+            "not printed; snippets read $TOWER_CATALOG_TOKEN",
+        ));
     }
-    output::write(&output::paragraph(
+    out.push_str(&output::paragraph(
         "These credentials are short-lived and intended for ad-hoc development use.",
     ));
-    output::newline();
+    out.push('\n');
 
     if !show_token {
-        output::newline();
-        output::header("Shell setup");
-        output::write(token_export_command(name, environment, mode, tower_url).as_str());
+        out.push('\n');
+        out.push_str(&header_line("Shell setup"));
+        out.push_str(token_export_command(name, environment, mode, tower_url).as_str());
     }
 
     for snippet in snippets(name, credentials, format, show_token) {
-        output::newline();
-        output::header(snippet.title);
-        output::write(snippet.body.as_str());
-        output::newline();
+        out.push('\n');
+        out.push_str(&header_line(snippet.title));
+        out.push_str(snippet.body.as_str());
+        if !snippet.body.ends_with('\n') {
+            out.push('\n');
+        }
+        out.push('\n');
     }
+
+    out
 }
 
 const PYICEBERG_TMPL: &str = include_str!("templates/pyiceberg.py.tmpl");
--- a/crates/tower-cmd/src/catalogs.rs
+++ b/crates/tower-cmd/src/catalogs.rs
@@ -1,8 +1,11 @@
-use clap::{value_parser, Arg, ArgMatches, Command};
+use clap::{value_parser, Arg, ArgAction, ArgMatches, Command};
 use colored::Colorize;
 use config::Config;
+use tower_api::models::{vend_catalog_credentials_body, CatalogCredentials};
 
 use crate::{api, output, util::cmd};
+
+const STORAGE_CATALOG_TYPE: &str = "tower-catalog";
 
 pub fn catalogs_cmd() -> Command {
     Command::new("catalogs")
@@ -17,14 +20,28 @@ pub fn catalogs_cmd() -> Command {
                         .default_value("default")
                         .value_parser(value_parser!(String))
                         .help("List catalogs in this environment")
-                        .action(clap::ArgAction::Set),
+                        .action(ArgAction::Set),
                 )
                 .arg(
                     Arg::new("all")
                         .short('a')
                         .long("all")
                         .help("List catalogs across all environments")
-                        .action(clap::ArgAction::SetTrue),
+                        .action(ArgAction::SetTrue),
+                )
+                .arg(
+                    Arg::new("type")
+                        .long("type")
+                        .value_parser(value_parser!(String))
+                        .help("Filter catalogs by type, e.g. tower-catalog")
+                        .action(ArgAction::Set),
+                )
+                .arg(
+                    Arg::new("storage")
+                        .long("storage")
+                        .help("List Tower-managed storage catalogs")
+                        .conflicts_with("type")
+                        .action(ArgAction::SetTrue),
                 )
                 .about("List all of your catalogs"),
         )
@@ -44,18 +61,68 @@ pub fn catalogs_cmd() -> Command {
                         .default_value("default")
                         .value_parser(value_parser!(String))
                         .help("Environment the catalog belongs to")
-                        .action(clap::ArgAction::Set),
+                        .action(ArgAction::Set),
                 )
                 .about("Show the details of a catalog, including its property names"),
+        )
+        .subcommand(
+            Command::new("credentials")
+                .arg(
+                    Arg::new("catalog_name")
+                        .value_parser(value_parser!(String))
+                        .index(1)
+                        .required(true)
+                        .help("Name of the catalog"),
+                )
+                .arg(
+                    Arg::new("environment")
+                        .short('e')
+                        .long("environment")
+                        .default_value("default")
+                        .value_parser(value_parser!(String))
+                        .help("Environment the catalog belongs to")
+                        .action(ArgAction::Set),
+                )
+                .arg(
+                    Arg::new("mode")
+                        .long("mode")
+                        .default_value("read")
+                        .value_parser(["read", "read-write"])
+                        .help("Credential access mode")
+                        .action(ArgAction::Set),
+                )
+                .arg(
+                    Arg::new("format")
+                        .long("format")
+                        .default_value("all")
+                        .value_parser(["all", "pyiceberg", "spark", "duckdb", "dbt"])
+                        .help("Snippet format to print")
+                        .action(ArgAction::Set),
+                )
+                .arg(
+                    Arg::new("show_token")
+                        .long("show-token")
+                        .help("Print the vended OAuth token in normal output")
+                        .action(ArgAction::SetTrue),
+                )
+                .about("Vend short-lived catalog credentials for external tools"),
         )
 }
 
 pub async fn do_list(config: Config, args: &ArgMatches) {
     let all = cmd::get_bool_flag(args, "all");
     let env = cmd::get_string_flag(args, "environment");
+    let catalog_type = if cmd::get_bool_flag(args, "storage") {
+        Some(STORAGE_CATALOG_TYPE)
+    } else {
+        args.get_one::<String>("type").map(String::as_str)
+    };
 
-    let catalogs =
-        output::with_spinner("Listing catalogs", api::list_catalogs(&config, &env, all)).await;
+    let catalogs = output::with_spinner(
+        "Listing catalogs",
+        api::list_catalogs(&config, &env, all, catalog_type),
+    )
+    .await;
 
     let headers = vec!["Name", "Type", "Environment"]
         .into_iter()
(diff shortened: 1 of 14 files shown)


konstantinoscs:
@coderabbitai re-trigger the failed workflow

crates/tower-cmd/src/catalogs.rs
```diff
@@ -26,6 +29,20 @@ pub fn catalogs_cmd() -> Command {
                         .help("List catalogs across all environments")
                         .action(clap::ArgAction::SetTrue),
                 )
+                .arg(
+                    Arg::new("type")
+                        .long("type")
+                        .value_parser(value_parser!(String))
+                        .help("Filter catalogs by type, e.g. tower-catalog")
+                        .action(ArgAction::Set),
```
socksy, line 39: now we have both `clap::ArgAction::Set` on line 23, and `ArgAction::Set` on this line - we should be consistent imo

crates/tower-cmd/src/catalogs.rs
```diff
@@ -119,9 +215,205 @@ pub async fn do_show(config: Config, args: &ArgMatches) {
...
+fn token_export_command(name: &str, environment: &str, mode: &str) -> String {
+    format!(
+        "export TOWER_CATALOG_TOKEN=\"$(tower --json catalogs credentials {name} --environment {environment} --mode {mode} | python3 -c 'import json,sys; print(json.load(sys.stdin)[\"credentials\"][\"oauth_token\"])')\"\n",
+        name = shell_quote(name),
+        environment = shell_quote(environment),
+        mode = shell_quote(mode),
+    )
+}
+
+fn print_credentials(
```
socksy: this seems like a function better suited for output.rs?

crates/tower-cmd/src/catalogs.rs
```diff
@@ -74,6 +141,35 @@ pub async fn do_list(config: Config, args: &ArgMatches) {
...
+        api::vend_catalog_credentials(&config, name, &env, parse_mode(mode)),
+    )
+    .await;
+
+    if output::get_output_mode().is_json() {
+        output::json(&response);
+        return;
+    }
+
+    print_credentials(name, &env, mode, &response.credentials, format, show_token);
```
socksy: doesn't seem to be any error logic if the credentials are failed to be fetched?

crates/tower-cmd/src/catalogs.rs
```diff
@@ -74,6 +141,35 @@ pub async fn do_list(config: Config, args: &ArgMatches) {
...
+        .unwrap_or("all");
+    let show_token = cmd::get_bool_flag(args, "show_token");
+
+    let response = output::with_spinner(
+        "Vending catalog credentials",
+        api::vend_catalog_credentials(&config, name, &env, parse_mode(mode)),
+    )
+    .await;
+
+    if output::get_output_mode().is_json() {
```
socksy: this is a smell that the `print_credentials` should be in `output.rs` and then takes care of whether to print out json inside of it to keep this irrelevant logic out of this file

crates/tower-cmd/src/catalogs.rs
```diff
@@ -119,9 +215,205 @@ pub async fn do_show(config: Config, args: &ArgMatches) {
...
+    format!("\"{}\"", value.replace('"', "\"\""))
+}
+
+fn token_export_command(name: &str, environment: &str, mode: &str) -> String {
+    format!(
+        "export TOWER_CATALOG_TOKEN=\"$(tower --json catalogs credentials {name} --environment {environment} --mode {mode} | python3 -c 'import json,sys; print(json.load(sys.stdin)[\"credentials\"][\"oauth_token\"])')\"\n",
+        name = shell_quote(name),
+        environment = shell_quote(environment),
+        mode = shell_quote(mode),
+    )
```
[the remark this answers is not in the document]
socksy: not sure this is actually important

konstantinoscs:
@coderabbitai you're not functionaing well today. I've adressed all of your feedback, check and resolve the comments

src/tower/_context.py
```diff
@@ -1,4 +1,37 @@
...
+        current = current.get(key)
+    return current if isinstance(current, str) and current else None
+
+
+def _read_session() -> dict[str, Any]:
+    session_path = Path.home() / ".config" / "tower" / "session.json"
+    try:
+        with session_path.open() as session_file:
+            session = json.load(session_file)
+    except (FileNotFoundError, OSError, json.JSONDecodeError, UnicodeDecodeError):
```
socksy: eh, why special case these errors if we're not differentiating between them?

src/tower/_context.py
```diff
@@ -32,11 +65,27 @@ def is_local(self) -> bool:
...
+        if tower_url is None or (tower_api_key is None and tower_jwt is None):
+            session = _read_session()
+
+        if tower_url is None:
+            tower_url = (
+                _get_session_value(session, "tower_url") or "https://api.tower.dev"
+            )
+
+        if tower_api_key is None and tower_jwt is None:
+            tower_jwt = _session_jwt(session)
```
socksy: Something seems super complicated here in all this logic getting session etc - IIRC in the cli we read from the env var, then the `.config/tower/ session.json` in turn, and state is just held by the fact that the latter can change (the former cannot, but `TOWER_API_KEY` isn't typically used in the scenario where that matters, which is a long running process — atm that's only the MCP server that e.g. needs to know that the team token changes when it changes team.

Here I feel it's already quite complicated (why do we have a context object with a build method? Doesn't feel pythonic at all to me) and on top of that I don't know why we need to support the sessions — is it expected that e.g. changing team in the CLI should somehow affect the local SDK? Would there be a scenario in which we want someone to setup this SDK and expect them to use the CLI session?

src/tower/_storage.py
```diff
@@ -0,0 +1,214 @@
...
+    try:
+        response = describe_default_catalog_api.sync_detailed(client=_env_client(ctx))
+        if response.status_code not in (HTTPStatus.OK, HTTPStatus.ACCEPTED):
+            return
+    except Exception:
+        # The following vend retry will surface the actionable backend/auth error.
+        return
+
+
+def _unwrap_vend_result(
```
socksy, line 136: why not just rely on the existing 404 error? needing to do this adds this function, and the need for `_is_not_found()` one-liner. Assuming there's an AI agent involved, can i maybe interest you in my `simpler` agent skill? https://github.com/socksy/agent-skills

src/tower/_tables.py
```diff
@@ -509,20 +598,27 @@ def create(self, schema: pa.Schema) -> Table:
...
             schema=convert_pyarrow_schema(schema),
         )
 
-        return Table(self._context, table)
+        return Table(
+            self._context,
+            table,
+            table_reference=self if self._tower_vended else None,
+            table_identifier=table_name,
+            catalog_mode=self._catalog_mode,
```
socksy, line 619: won't this always be "read-write" since you just ensured it above?

src/tower/_tables.py
```diff
@@ -509,20 +598,27 @@ def create(self, schema: pa.Schema) -> Table:
 
         namespace = namespace_or_default(self._namespace)
         table_name = make_table_name(self._name, namespace)
+        catalog = self._ensure_catalog_mode("read-write")
```
socksy, line 600: wouldn't it rather make sense to do this in the constructor function where we assign the catalog into the object? Seems weird to me to need to make the caller site do a sanitization when i would just expect to do things on the instance variables of the model? It doesn't seem there's any error in this path that you'd want to be lazily done

crates/tower-cmd/src/templates/dbeaver.txt.tmpl
```diff
@@ -0,0 +1,5 @@
+Catalog type: Iceberg REST
+URI: __TOWER_URI__
+Warehouse: __TOWER_WAREHOUSE__
+Authentication: Bearer token
+Token: __TOWER_TOKEN__
```
sammuti: I believe you need a scope field here

crates/tower-cmd/src/templates/duckdb.sql.tmpl
```diff
@@ -0,0 +1,9 @@
+duckdb <<SQL
```
sammuti, line 1: Duckdb can also do vended credential flow on the terminal
Something like
```
duckdb <<SQL
INSTALL httpfs;
LOAD httpfs;
INSTALL iceberg;
LOAD iceberg;

SET s3_region='eu-central-1';

CREATE OR REPLACE SECRET tower_cat (
  TYPE iceberg,
  CLIENT_ID '__TOWER_CLIENT_ID__',
  CLIENT_SECRET '__TOWER_CLIENT_SECRET__',
  OAUTH2_SCOPE 'PRINCIPAL_ROLE:ALL',
  ENDPOINT '__TOWER_URI__'
);

ATTACH '__TOWER_WAREHOUSE__' AS __TOWER_NAME__ (
  TYPE iceberg,
  SECRET tower_cat,
  DEFAULT_REGION 'eu-central-1',
  ACCESS_DELEGATION_MODE 'vended_credentials'
);
SQL
```

crates/tower-cmd/src/catalogs.rs
konstantinoscs: Yes!. Will do

crates/tower-cmd/src/catalogs.rs
konstantinoscs: The existing pattern is that command files format inline using output:: primitives — `do_show` does the exact same `if is_json() { output::json; return }` then `output::detail(...)`. Moving this one function out would make it inconsistent with `do_show`. That being said, maybe there's some value in DRY-ing the whole json output thing but outside the scope of this PR

crates/tower-cmd/src/catalogs.rs
konstantinoscs: Actually there is! `output::with_spinner` calls `tower_error_and_die(err, …)` on Err (output.rs), so a failed vend already dies with a proper Tower error — same as `do_list` actually.

crates/tower-cmd/src/catalogs.rs
konstantinoscs: Same as my comment above. It's an established pattern and while I agree that a refactor might be of value, I'd keep it out of this PR

src/tower/_context.py
konstantinoscs: Yeah I checked and this could be collapsed to `except (OSError, ValueError)`. Will do!

src/tower/_storage.py
konstantinoscs: You can interest my in your skill but that's a decision I've made heh. We need this because our provisioner is not lazy for the time being. So if the plan was to have a default catalog auto-provisioned this is needed.
I agree that the more proper fix for this behavior is server-side but that's a point of contention for the time being so I thought of putting it here.

Of course we could just not auto-provision and let this fail hard.
Open to discuss this

src/tower/_tables.py
konstantinoscs: There are two types of catalog here. if the catalog has credentials vended by tower, then yes this would be always a read-write. For any other catalog, this would be "read" although ignored. I think it's better for future expansion to not hardcode this.

crates/tower-cmd/src/templates/dbeaver.txt.tmpl
konstantinoscs: I looked into this more and I think the issue is a level higher than `scope`: DBeaver does not appear to have a native Apache Iceberg REST catalog connector, so this snippet does not map to a real DBeaver connection flow 😓 . The Iceberg REST bearer-token path itself does not require scope; scope is relevant for the OAuth/client-credentials flow, which we do not want to expose because the catalog principal credentials should stay inside Tower.

I’m leaning toward removing the DBeaver format for now unless you have a specific DBeaver/JDBC setup in mind that we should target and test.

crates/tower-cmd/src/templates/duckdb.sql.tmpl
konstantinoscs: This was deliberate. One of the core principles of our currest Storage Design is "master credentials never leave Tower" and using the functionality you described would require us to give back 2 "forbidden" secrets.
The snippet I posted here:

1. Does not leak any of the above mentioned
2. Is pretty much the exact same logic we use on our web app to attach to duck db :)

src/tower/_tables.py
konstantinoscs: I assume you mean only the new line and not the existing code. I think your instict is reasonable in general but we can't push this `ensure_catalog_mode()` into the constructor and that's a direct consequence of our decision to allow read-only access (and in general support RBAC):
1.  A TableReference is read by default and shared between reads and writes — the same tables("x") is used for both `.load().read()` and `.create()/.insert()`.
2. If `__init__` did `_ensure_catalog_mode("read-write")`, every `tables("x")` would vend read-write creds up front. That breaks the least-privilege model we deliberately decided on (funny enough, coderabbit had a comment exactly about that) : a read-only principal — (a service account without catalogs:data:write) — would 403 at construction even when it only wants to read, and read-only flows would over-acquire write credentials.

So we actively do want this lazy: the read-write escalation (and its possible failure) should happen only when a write actually occurs. There's a "fail-fast" cosst beacause you learn you can't write at the moment you try to write but it's the price of not forcing write-perms onto read-only usage.

src/tower/_context.py
konstantinoscs: I think you're right on this one. I tried to follow a decision that seems quite insignificant in retrospect: "a `tables()` call works out of the box after a `tower login`". To do that, I made the SDK session-aware but I agree that an SDK should be uninformed of the state. The cost we'll pay is the explicit setting of `TOWER_API_KEY` but it looks like a much cleaner design to me. Thanks for this comment!!

crates/tower-cmd/src/catalogs.rs
socksy: Eh I don't think that's an established pattern really? If it weren't for Ctrl-C then this function wouldn't be public in the first place. Basically knowing whether you're printing in JSON mode should be completely outside the remit of code not in `output.rs`. The other case is maybe arguable (eh) but this one isn't IMO

crates/tower-cmd/src/templates/duckdb.sql.tmpl
sammuti: This does not require master credentials to leave it is just user specific principle and secrets which users receive when using plane polaris catalogs like Snowflake. I personally do this my self when connecting with with rest catalogs.

crates/tower-cmd/src/catalogs.rs
konstantinoscs: Hmm ok I created a helper `render_or_json()` that keeps the "am I in json" logic in output.rs

src/tower/_context.py
konstantinoscs: Done!

crates/tower-cmd/src/templates/duckdb.sql.tmpl
konstantinoscs: Yeah that's what I actually mean by master credentials - master credentials for the user. We don't let those leave the Tower control plane, just the minted vended credentials we return from the API

src/tower/_context.py
konstantinoscs: Ok after all I've completely removed these additions since we're not doing session fallback

crates/tower-cmd/src/templates/duckdb.sql.tmpl
sammuti: Hmm seems we are breaking the iceberg vended credentials standard and inventing a new way to create tokens which requires user to hard code a long lived token or write their own script always instead of standard iceberg client tooling which support the vended credentials flow which works with X-Iceberg-Access-Delegation header

konstantinoscs: I think there are two different concerns mixed together here.

We are not bypassing the Iceberg REST access-delegation flow. The DuckDB snippet uses bearer-token auth into the REST catalog, and the attach still uses `ACCESS_DELEGATION_MODE 'vended_credentials'` / `X-Iceberg-Access-Delegation: vended-credentials` for storage creds. I’ll make that explicit in the snippet so it’s obvious.

The reason we don’t emit `CLIENT_ID` / `CLIENT_SECRET` is Tower V1’s principal model. V1 does not have per-user catalog principals or per-user catalog RBAC. Each Tower-managed catalog has shared Tower principals (`tower_ro_*` / `tower_rw_*`), and Tower enforces team membership before minting short-lived tokens. So the only client secret we could give to DuckDB would be a shared catalog principal secret, not a user-specific secret.

That would be a worse security posture:
- it is long-lived;
- it is shared across the team;
- it cannot be revoked per user;
- rotation would affect everyone;
- for RW it would effectively expose the shared write principal.

So this PR intentionally keeps those principal credentials inside Tower and only returns short-lived, per-call bearer tokens after Tower auth. A more durable external-tool setup probably needs V2 per-user/per-service-account catalog principals, not exporting the shared V1 principals.

socksy (approved):

PR #313, opened 2026-07-01
fix: correctly detect which catalogs we should vend credentials for

konstantinoscs:
The problem was that we try to vend credentials for all tables in the `tables()` call so it messed up s3 tables. This fixes that

### DIFF
--- a/src/tower/_storage.py
+++ b/src/tower/_storage.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import hashlib
+import logging
 import time
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
@@ -12,11 +13,13 @@ from ._context import TowerContext
 from .tower_api_client.api.default import (
     describe_default_catalog as describe_default_catalog_api,
 )
+from .tower_api_client.api.default import describe_catalog as describe_catalog_api
 from .tower_api_client.api.default import (
     vend_catalog_credentials as vend_catalog_credentials_api,
 )
 from .tower_api_client.models import (
     CatalogCredentials,
+    DescribeCatalogResponse,
     ErrorModel,
     VendCatalogCredentialsBody,
     VendCatalogCredentialsBodyMode,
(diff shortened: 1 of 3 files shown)


src/tower/_storage.py
```diff
@@ -123,6 +127,34 @@ def _vend_catalog_credentials(
...
+        return _catalog_type_cache[cache_key]
+
+    try:
+        result = describe_catalog_api.sync(
+            name=name,
+            client=_env_client(ctx),
+            environment=environment,
+        )
+    except Exception:
+        return None
```
bradhe, line 156: Do we want to get this out for diagnostic purposes somehow? I'm not really sure how this call could fail.

src/tower/_tables.py
```diff
@@ -60,6 +66,45 @@ def _load_tower_catalog(
...
+    prefix = _pyiceberg_catalog_env_prefix(name)
+    return any(key.upper().startswith(prefix) for key in os.environ)
+
+
+def _should_vend_tower_credentials(
+    ctx: TowerContext,
+    name: str,
+    environment: str,
+    tower_credentials: Optional[bool],
+) -> bool:
```
socksy, line 92: why take all these args lol just to do a `not None` check on a single arg? i'd just inline this one line 837 instead

src/tower/_storage.py
konstantinoscs: I'm not sure what you mean? If we want to report it?

src/tower/_tables.py
socksy: oops my wrapping made it look like this function ended after the first return, ignore this

src/tower/_storage.py
socksy: like if you hit this exception let's log something or die louder rather than swallow it and pretend nothing went wrong, since it shouldn't break in usual usage

konstantinoscs: I added a log here. But we *should not die loudly*:

We use this function to answer: “Is this string catalog a Tower-managed tower-catalog, so we should vend Tower credentials, or is it BYO/S3 Tables, so we should use PyIceberg config?”
If the probe fails, we still have a valid fallback path: check for PyIceberg config and then call load_catalog(). That’s especially important for BYO/S3 Tables customers, because their catalog may work perfectly through runner-injected PYICEBERG_CATALOG__... config even if Tower’s describe API is unavailable, unauthorized, stale, or temporarily failing.

socksy (approved):

PR #314, opened 2026-07-03
Correctly create wheels for linux, musllinux, and musllinux-cross

jo-sm:
This PR fixes wheel artifact creation for `linux`, `musllinux`, and `musllinux-cross`, so that all expected Python versions are correctly published.


konstantinoscs (approved):
LGTM.
One thing is that we could/should build is a post-publish job that checks that the wheels/ dir contains all the files that should be there. (it's my understanding of the original problem that we were skipping some of them)

PR #321, opened 2026-07-06
Use ARM runners

jo-sm:
This PR updates the `linux` and `musllinux` jobs to use ARM runners directly, rather than doing cross compilation.

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -42,6 +42,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           submodules: recursive
+          persist-credentials: false
       - uses: actions-rust-lang/setup-rust-toolchain@v1.11.0
       - uses: actions/setup-python@v6
         with:
@@ -72,6 +73,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           submodules: recursive
+          persist-credentials: false
       - name: "Install fresh upstream rustup"
         # The macos-15-arm64 runner image (20260511.0048+) ships a broken
         # Homebrew rustup whose bundled rustup-init rejects standard argv
@@ -117,6 +119,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           submodules: recursive
+          persist-credentials: false
       - name: "Install fresh upstream rustup"
         # The macos-15-arm64 runner image (20260511.0048+) ships a broken
         # Homebrew rustup whose bundled rustup-init rejects standard argv
@@ -170,6 +173,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           submodules: recursive
+          persist-credentials: false
       - uses: actions-rust-lang/setup-rust-toolchain@v1.11.0
       - uses: actions/setup-python@v6
         with:
@@ -204,6 +208,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           submodules: recursive
+          persist-credentials: false
       - uses: actions-rust-lang/setup-rust-toolchain@v1.11.0
       - uses: actions/setup-python@v6
         with:
@@ -212,7 +217,7 @@ jobs:
         uses: PyO3/maturin-action@v1.50.1
         with:
           target: ${{ matrix.platform.target }}
-          manylinux: auto
+          manylinux: 2_28
           args: --release --locked --out dist -i python${{ matrix.python-version }}
           before-script-linux: |
             # If we're running on rhel centos, install needed packages.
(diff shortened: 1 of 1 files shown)


.github/workflows/build-binaries.yml
```diff
@@ -237,60 +236,44 @@ jobs:
...
-            ln -s -f /usr/bin/${{ format('pip{0}', matrix.python-version) }} /usr/bin/pip3
-            # Workaround for QEMU bug when emulating 32-bit on 64-bit hosts
-            # See: https://github.com/docker/buildx/issues/395
-            # Use /tmp for CARGO_HOME (often tmpfs, which avoids the QEMU filesystem bug)
-            export CARGO_HOME=/tmp/cargo-home
-            python3 -m venv /tmp/venv
-            . /tmp/venv/bin/activate
-            pip3 install dist/*.whl --force-reinstall
-            ${{ env.EXECUTABLE_NAME }} --help
+          manylinux: auto
```
konstantinoscs: We should put this back to 2_28. I dug a bit on what auto means and it basically says "pick the lowest compatible/default manylinux setup". The logs for the CI on this PR show that this effectively downgraded us to 2_17. We moved to 2_28 on purpose at #277 for very good reasons so we should keep this

.github/workflows/build-binaries.yml
```diff
@@ -237,60 +236,44 @@ jobs:
...
-            arch: aarch64
-            # see https://github.com/astral-sh/ruff/issues/3791
-            # and https://github.com/gnzlbg/jemallocator/issues/170#issuecomment-1503228963
-            maturin_docker_options: -e JEMALLOC_SYS_WITH_LG_PAGE=16 -e PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1
         python-version: ["3.12", "3.13", "3.14"]
     steps:
       - uses: actions/checkout@v6
         with:
           submodules: recursive
-          persist-credentials: false
```
konstantinoscs, line 256: We need to keep this. persist-credentials runs with true by default and we don't want it here as subsequent steps run actual code and having a GITHUB token available to them is.... not good.

.github/workflows/build-binaries.yml
```diff
@@ -353,24 +335,23 @@ jobs:
       - uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Install OpenSSL
```
konstantinoscs: This step is not needed. It installs OpenSSL on the host but the actual job is run inside a musl Docker environment

konstantinoscs (changes requested):
There are a couple of critical things we should fix here. Otherwise looks good to me

.github/workflows/build-binaries.yml
```diff
@@ -204,16 +208,16 @@ jobs:
...
       - uses: actions-rust-lang/setup-rust-toolchain@v1.11.0
       - uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python-version }}
-          architecture: x64
       - name: Build wheel
         uses: PyO3/maturin-action@v1.50.1
         with:
           target: ${{ matrix.platform.target }}
-          manylinux: auto
```
konstantinoscs, line 216: I see you moved the other jobs to 2_28 as well. I guess because of consistency. That's ok for me but the cost we're paying is that for those architectures, we're dropping older glibc users (e.g. someone on Ubuntu 18.04).

For the arm runner we need 2_28 because of the bug we have (again, in #277).
We don't techincally need the same for x64 and we could theoretically support older glibc users but I think being consistent is better for maintenability here. So I'd say leave it, I just wanted to point it out

.github/workflows/build-binaries.yml
```diff
@@ -262,35 +263,23 @@ jobs:
...
-            export CARGO_HOME=/tmp/cargo-home
-            python3 -m venv /tmp/venv
-            . /tmp/venv/bin/activate
-            pip3 install dist/*.whl --force-reinstall
-            ${{ env.EXECUTABLE_NAME }} --help
+          args: --release --locked --out dist -i python${{ matrix.python-version }}
+          before-script-linux: |
+            # If we're running on rhel centos, install needed packages.
+            if command -v yum &> /dev/null; then
+                yum update -y && yum install -y perl-core openssl openssl-devel pkgconfig libatomic
```
konstantinoscs: I also don't think we need this because we don't use any native-tls anymore. We're on ring (as tls provider) + rustls.

Let's drop this

konstantinoscs (approved):

.github/workflows/build-binaries.yml
jo-sm: Yeah I'd rather be consistent unless there's truly a need to have `auto` here. If it ends up causing issues for someone on an older version of Linux we can definitely revert this or pin to an older version.

PR #322, opened 2026-07-06
refactor(cli): replace the output globals with an explicit `Out`

socksy:
Where CLI output went was controlled by two mutable globals: an `OUTPUT_MODE` and a `CURRENT_SENDER` that the MCP server set and cleared around each tool call. Every `output::*` function read them to choose between stdout, JSON, and the MCP peer. So a command never said where its output went. That was decided elsewhere, out of sight, and the MCP path had to juggle a global sender to capture anything a command printed.

This passes the destination in as a value instead. Commands take an `Out`, and the writing functions become methods on it. The CLI builds one over stdout. For an MCP tool, the `Out` wraps a writer that forwards each line to that call's channel. The sender and the notification flag live on the value now, not in globals.

Most of the ~17-file diff is the mechanical `output::foo(x)` → `out.foo(x)` sweep.

`Out` is cloneable so the spawned log and status monitors can share the writer, which is why it sits behind an `Arc<Mutex>`. Tests, build, and fmt pass.

### DIFF
--- a/crates/tower-cmd/src/util/git.rs
+++ b/crates/tower-cmd/src/util/git.rs
@@ -13,12 +13,18 @@ use tower_telemetry::debug;
 /// This is used to auto-populate the `X-Tower-Idempotency-Key` header on deploy.
 pub fn clean_head_sha(dir: &Path) -> Option<String> {
     if !is_inside_work_tree(dir) {
-        debug!("{:?} is not inside a git worktree; skipping idempotency key", dir);
+        debug!(
+            "{:?} is not inside a git worktree; skipping idempotency key",
+            dir
+        );
         return None;
     }
 
     if is_dirty(dir) {
-        debug!("git worktree at {:?} is dirty; omitting idempotency key", dir);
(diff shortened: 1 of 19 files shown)


crates/tower-cmd/src/util/git.rs
```diff
@@ -13,12 +13,18 @@ use tower_telemetry::debug;
 /// This is used to auto-populate the `X-Tower-Idempotency-Key` header on deploy.
 pub fn clean_head_sha(dir: &Path) -> Option<String> {
     if !is_inside_work_tree(dir) {
-        debug!("{:?} is not inside a git worktree; skipping idempotency key", dir);
+        debug!(
```
socksy, line 16: unrelated but cargo fmt was failing on this file

bradhe (approved):

konstantinoscs (approved):
LGTM, I wonder if there's a way to actually test the concurrent call scenario (and that the channels never cross-talk

PR #323, opened 2026-07-07
Add Rust cache to all jobs

jo-sm:
This PR adds the `Swatinem/rust-cache` action to all of the jobs, which improves workflow times by approx. 1 minute.


socksy (approved):
yay

PR #324, opened 2026-07-11
feat: add beta label to CLI output

konstantinoscs:
Adding a resuable "beta feature" component that prints a beta label/warning when a beta feature is used.
Also, using the above mechanism for the Storage product commands ahead of the Preview/Beta launch.

e.g.
image

## Summary by CodeRabbit

* **New Features**
  * Added beta labeling and one-time notifications for Storage catalog functionality.
  * Storage-related CLI help now clearly identifies beta status and may include documentation links.
  * Beta notices appear only in suitable interactive terminal contexts and are shown once per feature.

* **Bug Fixes**
  * Prevented duplicate beta notices, including during concurrent CLI usage.

* **Tests**
  * Added coverage for notice formatting, validation, concurrency, persistence errors, and CLI help output.

### DIFF
--- a/crates/tower-cmd/src/beta.rs
+++ b/crates/tower-cmd/src/beta.rs
@@ -1,0 +1,103 @@
+use std::io::{self, IsTerminal};
+
+use tower_telemetry::debug;
+
+use crate::output::{self, OutputMode};
+
+pub(crate) struct BetaFeature {
+    id: &'static str,
+    message: &'static str,
+    docs_url: Option<&'static str>,
+}
+
+impl BetaFeature {
+    pub fn short_about(&self, description: &str) -> String {
+        format!("{description} [beta]")
+    }
+
+    pub fn notice(&self) -> String {
+        match self.docs_url {
+            Some(url) => format!("{} Learn more: {url}", self.message),
+            None => self.message.to_string(),
+        }
+    }
+}
+
+pub(crate) const STORAGE_BETA_MESSAGE: &str = "Tower Storage is in beta. Core functionality is stable, but some featues and interfaces might change before general availability.";
+
+pub(crate) const STORAGE: BetaFeature = BetaFeature {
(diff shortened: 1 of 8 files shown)


crates/tower-cmd/src/beta.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    pub fn notice(&self) -> String {
+        match self.docs_url {
+            Some(url) => format!("{} Learn more: {url}", self.message),
+            None => self.message.to_string(),
+        }
+    }
+}
+
+pub(crate) const STORAGE_BETA_MESSAGE: &str = "Tower Storage is in beta. Core functionality is stable, but its API contract and CLI interface may evolve before general availability.";
```
bradhe: Let's generalize this a little bit. 

```suggestion
pub(crate) const STORAGE_BETA_MESSAGE: &str = "Tower Storage is in beta. Some features and functionality might change.";
```

crates/tower-cmd/src/beta.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+}
+
+fn should_notify(
+    output_mode: OutputMode,
+    stdout_is_terminal: bool,
+    stderr_is_terminal: bool,
+) -> bool {
+    output_mode.is_normal() && stdout_is_terminal && stderr_is_terminal
+}
```
bradhe, line 56: Can you add a check to this that is if the user was recently notified? I don't think we need to do this on every invokation.

bradhe (approved):

crates/tower-cmd/src/beta.rs
konstantinoscs: if you check, this function is called from inside `notify_once()` that has the semantics you're looking for. We atomically persist a marker which will skip this function on subsequent calls.

PR #328, opened 2026-07-21
Add query support to local MCP server

bradhe:
Title says it all. Should make our output a little nicer/easier to use in some cases!

### DIFF
--- a/crates/tower-cmd/src/mcp.rs
+++ b/crates/tower-cmd/src/mcp.rs
@@ -199,6 +199,18 @@ struct ShowCatalogRequest {
     environment: Option<String>,
 }
 
+#[derive(Debug, Deserialize, JsonSchema)]
+struct QueryCatalogRequest {
+    /// Name of the Tower-managed storage catalog to query
+    name: String,
+    /// A single read-only SQL statement. Tables must be fully qualified as
+    /// <catalog>.<namespace>.<table> (there is no default schema). Writes, DDL,
+    /// and multiple statements are rejected.
+    sql: String,
+    /// The environment the catalog belongs to (defaults to "default")
+    environment: Option<String>,
+}
+
 pub fn mcp_cmd() -> Command {
     Command::new("mcp-server")
         .about("Runs an MCP server for LLM interaction")
@@ -719,7 +731,9 @@ impl TowerService {
         }
     }
 
-    #[tool(description = "Show details for a catalog, including its property names")]
+    #[tool(
+        description = "Show a catalog's details: its property names and, for Tower-managed storage catalogs, the namespaces and tables you can query."
+    )]
     async fn tower_catalogs_show(
         &self,
         Parameters(request): Parameters<ShowCatalogRequest>,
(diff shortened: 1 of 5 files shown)


konstantinoscs (changes requested):
I think we need to reconsider this. Basically this PR allows agents to run arbitary SQL that gets directly forwarded to duckDB. 

DuckDB SQL can read local files, access the network, load extensions, and consume unbounded resources. DuckDB’s own security guidance says to treat untrusted SQL like Bash or Python code and recommends OS/container/WASM isolation. See:
https://duckdb.org/docs/current/operations_manual/securing_duckdb/overview

To be fair, this happens already in the human operated mode and probably this should be changed as well but an agent increases the security risk manyfold.

Btw, in the monorepo we avoid this by rusing WASM and adding read-only and single-statement "gates".

I think we need to do some sandboxing here

crates/tower-cmd/src/mcp.rs
```diff
@@ -750,6 +763,48 @@ impl TowerService {
         }
     }
 
+    #[tool(
+        description = "Run a read-only SQL query against a Tower-managed storage (Iceberg) catalog. Tables must be fully qualified as <catalog>.<namespace>.<table>. Use tower_catalogs_show first to discover namespaces and tables. Returns the result columns and rows as JSON."
```
konstantinoscs: That MCP method only returns catalog properties so iiuc, it won't work. Have you tried this? Probably should add tests as well

konstantinoscs: Also, here we claim a "read-only" query but me understanding is that if something provides "read-write" credentials it will work. Agent mutations should be guarded imo

crates/tower-cmd/src/mcp.rs
```diff
@@ -750,6 +763,48 @@ impl TowerService {
...
+        let sql = request.sql.trim().to_string();
+        if sql.is_empty() {
+            return Self::text_error("No SQL statement provided.".to_string());
+        }
+        let write = request.write.unwrap_or(false);
+
+        match crate::catalogs::query_catalog(&self.config, &request.name, environment, sql, write)
+            .await
+        {
+            Ok(result) => {
```
konstantinoscs: Results are unbounded right? That could lead into memory exhaustion. I can foresee an agent trying to do a query for millions of rows. We should either bound the result or return some `truncated` marker. In the web ui we have a limit of 100k rows.

crates/tower-cmd/src/mcp.rs
```diff
@@ -750,6 +763,48 @@ impl TowerService {
...
+                            result
+                                .columns
+                                .iter()
+                                .cloned()
+                                .zip(row.iter().cloned())
+                                .collect(),
+                        )
+                    })
+                    .collect();
+                Self::json_success(json!({
```
konstantinoscs: I think here you are silently overwriting duplicate column names. A `join` query could potentially produce duplicate names which are both valid but here the last one will win. In the UI we are using "positional arrays" exactly for that reason (see catalogService.ts)

konstantinoscs (changes requested):
My previous security concern remains

crates/tower-cmd/src/mcp.rs
bradhe: We can easily disallow read-write credentials to the catalog, you're right that we should come up with a way to sandbox the DuckDB session overall.

crates/tower-cmd/src/mcp.rs
bradhe: This is a good suggestion.

crates/tower-cmd/src/mcp.rs
bradhe: Good catch — the MCP `tower_catalogs_show` only returned properties, so an agent couldn't actually discover namespaces/tables from it. It now lists tables too, via a shared `list_catalog_tables` helper (the same path the CLI `show` uses), returning `tables: [{namespace, table}]` (plus `tables_error` when the listing fails, without failing the whole call). Added unit tests for the query gates, row truncation, and session hardening. Fixed in aeaf9cb.

bradhe: Done — removed the `write` option from the MCP tool entirely. It always vends read-only credentials and attaches `READ_ONLY`, and there's now a leading-keyword gate that rejects write/DDL statements as defence-in-depth. `--write` stays only on the human CLI path.

crates/tower-cmd/src/mcp.rs
bradhe: Bounded — agent queries are now capped at 1000 rows and the response carries `"truncated": true` when more matched. I went below the UI's 100k because these rows land directly in a model's context; it's a single `AGENT_MAX_ROWS` const if we want to tune it.

crates/tower-cmd/src/mcp.rs
bradhe: Fixed — the MCP response now returns positional row arrays (`rows: [[...]]`, one value per column in column order) like `catalogService.ts`, so duplicate names from a join aren't collapsed.

konstantinoscs (approved):
My main concerns about security are solved, thanks!

There are a few nits regarding resource consumption constraints that I think remain. I tested them locally with the help of Sol. You can have your agent take a stab at them but feel free to merge afterwards:

1. Nested JSON bypasses the 1 MiB response cap.
The limiter undercounts arrays, strings requiring JSON escaping, separators, and pretty-printing. The entire nested DuckDB value is also materialized before the estimate is checked.
A local probe with:
`SELECT list_transform(range(10000000), x -> '') AS xs`
was allowed and returned one row with truncated:false, but the MCP-equivalent serialized response was 120,000,115 bytes.

2. DuckDB errors bypass the result ceiling entirely.
`query_catalog_for_agent` forwards `err.to_string() `directly through MCP.
This allowed query:
`SELECT CAST(repeat('x', 2000000) AS INTEGER)`
generated a 2,000,126-byte error containing the offending value. The same pattern can use an aggregate over a catalog column, bypassing the intended output cap through the error channel.
Return a bounded, stable error to MCP and retain detailed errors only in protected logs. Token redaction is present and correct, but it does not bound the message.

3. Parallel calls multiply the per-query resource limits.
Every request starts a new blocking DuckDB session at `catalogs.rs` line 690. MCP requests are dispatched concurrently, and there is no service-wide semaphore in TowerService.
Each session may consume 1 GiB RAM and 2 GiB temporary storage, so a handful of concurrent agent calls can exhaust the machine. Cancelling the MCP future also does not terminate an already-running spawn_blocking task.
This differs from the storage web product, which explicitly serializes every engine operation through a single promise chain in `catalogService.ts` (line 171). Add a process-wide shared concurrency budget and cancellation-aware execution.

4. Catalog discovery remains unbounded.
`list_catalog_tables` runs `SHOW ALL TABLES` with `Limits::none()`, after which `tower_catalogs_show` serializes every table into model context.
Give the MCP discovery path pagination or row/byte limits and return tables_truncated or a cursor.

5. My understanding of reading the new tests you've added is that they won't run in the CI? Is that intentional?

PR #331, opened 2026-07-24
feat: tower-duckdb crate — sandboxed catalog queries and adversarial tests

bradhe:
## What this is

This PR adds a `tower-duckdb` crate that owns Tower's usage of DuckDB, and wires its query sandbox into `tower catalogs query`. It carries Konstantinos's review feedback from #328 (read-only enforcement, session sandboxing, bounded results) forward as a reusable, tested foundation. The MCP `tower_catalogs_query` tool that also consumes the crate lands in a follow-up on #328.

## Why it matters

Once an agent can issue SQL against a customer's catalog, the query text is untrusted input running on the customer's machine with their credentials. Read-only credentials are not enough on their own: a query can still read the local filesystem through `read_csv('/etc/passwd')`, reach internal network endpoints through table functions, load an arbitrary extension, or smuggle a second statement past a naive single-statement check. This PR closes that surface, puts the whole DuckDB integration in one crate so there is a single place to reason about it, and pins the invariants down with tests so a future change cannot quietly reopen them.

### DIFF
--- a/crates/tower-cmd/src/catalogs.rs
+++ b/crates/tower-cmd/src/catalogs.rs
@@ -7,168 +7,12 @@ use std::time::{Duration, Instant};
 use tower_api::models::{
     vend_catalog_credentials_body, CatalogCredentials, DescribeCatalogResponse,
 };
+use tower_duckdb::{guard, params, run_query, QueryResult, Session};
 use tower_telemetry::debug;
 
 use crate::{api, beta, output, util::cmd};
 
 const STORAGE_CATALOG_TYPE: &str = "tower-catalog";
-
-/// Query-sandbox primitives for untrusted (agent) callers. These are the gates
-/// and the DuckDB session lockdown a future MCP `tower_catalogs_query` tool will
-/// apply to agent-issued SQL: reject write/DDL and multi-statement input, cap
-/// the row count, and lock the session down so a query cannot read the local
-/// filesystem, load community extensions, or unwind the settings. Nothing wires
-/// these into the query path yet, so `dead_code` is allowed here; the
-/// adversarial test suite exercises them directly.
-#[allow(dead_code)]
-mod sandbox {
-    /// Row cap for agent-issued queries. Rows past this are dropped and the
-    /// caller is told the set was truncated, so a model cannot pull an unbounded
-    /// table into memory or its context.
-    pub(super) const AGENT_MAX_ROWS: usize = 1_000;
-
-    /// Leading keywords that write data or schema, or repoint the session. An
-    /// agent query starting with one of these is refused, as defence in depth on
-    /// top of the read-only credentials it is given.
-    const WRITE_LEADING_KEYWORDS: &[&str] = &[
-        "insert", "update", "delete", "merge", "create", "drop", "alter", "truncate", "replace",
-        "copy", "attach", "detach",
-    ];
-
-    /// Statements that lock an attached DuckDB session down before an untrusted
-    /// query runs: no local-filesystem access (so `read_csv('/etc/passwd')` and
-    /// friends fail), no community extensions, and a configuration lock so the
-    /// query cannot unwind any of it. These run after the catalog is attached,
-    /// which is what installs extensions and reaches the network. Network access
-    /// stays on because httpfs is how the attached Iceberg catalog reads its
-    /// data, so this narrows but does not eliminate the query surface.
-    pub(super) fn agent_hardening_statements() -> Vec<String> {
-        vec![
-            "SET disabled_filesystems = 'LocalFileSystem'".to_string(),
-            "SET allow_community_extensions = false".to_string(),
-            "SET lock_configuration = true".to_string(),
-        ]
-    }
-
-    /// The leading SQL keyword, lowercased, after skipping leading whitespace and
-    /// `--` / `/* */` comments.
-    pub(super) fn first_sql_keyword(sql: &str) -> String {
-        let mut s = sql.trim_start();
-        loop {
-            if let Some(rest) = s.strip_prefix("--") {
-                match rest.find('\n') {
-                    Some(nl) => s = rest[nl + 1..].trim_start(),
-                    None => return String::new(),
-                }
-            } else if let Some(rest) = s.strip_prefix("/*") {
-                match rest.find("*/") {
-                    Some(end) => s = rest[end + 2..].trim_start(),
-                    None => return String::new(),
-                }
-            } else {
-                break;
-            }
-        }
-        s.chars()
-            .take_while(|c| c.is_ascii_alphabetic())
-            .collect::<String>()
-            .to_lowercase()
-    }
-
-    /// True when `sql` starts with a write/DDL keyword.
-    pub(super) fn is_write_statement(sql: &str) -> bool {
-        WRITE_LEADING_KEYWORDS.contains(&first_sql_keyword(sql).as_str())
-    }
-
-    /// True when `sql` holds more than one statement. A `;` inside a string
-    /// literal or comment is data, not a separator, so those spans are skipped.
-    /// This gate matters because duckdb-rs `prepare` runs every statement but the
-    /// last as a side effect, so unguarded multi-statement SQL would execute its
-    /// leading statements even though only the final one is returned.
-    pub(super) fn contains_multiple_statements(sql: &str) -> bool {
-        #[derive(PartialEq)]
-        enum State {
-            Normal,
-            Single,
-            Double,
-            Line,
-            Block,
-        }
-
-        let mut state = State::Normal;
-        let mut statements = 0usize;
-        let mut current_has_content = false;
-        let mut chars = sql.chars().peekable();
-
-        while let Some(c) = chars.next() {
-            match state {
-                State::Normal => match c {
-                    '\'' => {
-                        state = State::Single;
-                        current_has_content = true;
-                    }
-                    '"' => {
-                        state = State::Double;
-                        current_has_content = true;
-                    }
-                    '-' if chars.peek() == Some(&'-') => {
-                        chars.next();
-                        state = State::Line;
-                    }
-                    '/' if chars.peek() == Some(&'*') => {
-                        chars.next();
-                        state = State::Block;
-                    }
-                    ';' => {
-                        if current_has_content {
-                            statements += 1;
-                            if statements > 1 {
-                                return true;
-                            }
-                        }
-                        current_has_content = false;
-                    }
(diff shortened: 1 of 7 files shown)


crates/tower-duckdb/src/guard.rs
```diff
@@ -0,0 +1,135 @@
...
+/// Leading keywords that write data or schema, or repoint the session. A query
+/// starting with one of these is refused before it runs.
+const WRITE_LEADING_KEYWORDS: &[&str] = &[
+    "insert", "update", "delete", "merge", "create", "drop", "alter", "truncate", "replace",
+    "copy", "attach", "detach",
+];
+
+/// The leading SQL keyword, lowercased, after skipping leading whitespace and
+/// `--` / `/* */` comments.
+pub fn first_keyword(sql: &str) -> String {
```
konstantinoscs: We should include /r here as it's a valid ending to a comment for dubkDB SQL. A payload beginning -- comment\r... ca look empty to both gates but be parsed as executable SQL by DuckDB. 

More broadly, checking only the first keyword is not a safe read-only policy: statements beginning with WITH, for example, can still contain mutations. 
I think this should use DuckDB’s parser/tokenizer and an allowlist of parsed statement types, not a keyword denylist

crates/tower-cmd/src/catalogs.rs
```diff
@@ -497,7 +513,20 @@ async fn execute_catalog_query(
 
     let token = response.credentials.oauth_token.clone();
     let setup = attach_statements(name, &response.credentials, mode);
-    let result = tokio::task::spawn_blocking(move || run_duckdb_query(&setup, &sql, [])).await;
+    // Read mode is the sandboxed path: lock the session down after attach and cap
+    // the rows, so the query cannot read the host, escape the config, or pull an
+    // unbounded table back. Write mode is trusted and runs the setup as-is.
+    let harden = !write;
+    let max_rows = (!write).then_some(guard::AGENT_MAX_ROWS);
```
konstantinoscs: Btw we're "limiting" rows in memory so we don't escape the resource exhaustion. The queries will still fetch all the rows in memory and then truncate the result so I think at this point the damage has already been done. Not the end of the world, just flagging it out.

konstantinoscs (changes requested):
The SQL injection please. The rest is ok!

crates/tower-duckdb/src/guard.rs
bradhe: Good catch, and agreed on the broader point. Fixed in 7d7d2af38: the gate now runs the SQL through DuckDB's own parser via `json_serialize_sql`, which parses without executing and serializes only SELECT statements, erroring on everything else. `classify_read_only` returns `Allowed` only for exactly one SELECT; writes, DDL, PRAGMA/SET, multi-statement, and unparseable input are all refused fail-closed. So it's now an allowlist of what the executor will actually run, not a first-keyword denylist.

Both of your examples are covered by new tests: `-- harmless\rDROP TABLE runs` parses as a DROP and is rejected, and `WITH x AS (DELETE FROM t RETURNING *) SELECT * FROM x` is rejected as a parser error while `WITH x AS (SELECT 1) SELECT ...` is still allowed. The SQL is bound as a parameter, never spliced into the parser query.

crates/tower-cmd/src/catalogs.rs
bradhe: Agreed, and thanks for flagging it. The cap is a context/output bound, not a resource-exhaustion defense, and I'll keep it as the former.

One nuance: the read loop breaks at the cap and stops pulling, so for a streamable plan (a scan/filter) DuckDB stops producing once we stop reading, which does bound what the client pulls and holds. What it can't help is an inherently materializing plan (an ORDER BY or aggregate over a huge input), where DuckDB builds the full result internally before we see the first row, so the memory is spent regardless of any client-side cap. I don't think there's a clean fix for that from our side short of refusing/altering the user's query, so I'd rather leave the query semantics intact and treat the cap as "don't dump 10M rows into a terminal or a model's context." Happy to revisit if we want a hard resource ceiling later (DuckDB `memory_limit` / `max_temp_directory_size` would be the lever, not row count).

crates/tower-duckdb/src/guard.rs
```diff
@@ -0,0 +1,478 @@
...
+/// `nextval`/`currval` mutate sequence state, so they are writes wearing a
+/// SELECT's clothes. `query`/`query_table` hand a string to DuckDB's execution
+/// pipeline, which turns this gate's allowlist into an execution surface. None of
+/// them have a legitimate use in a catalog data query.
+///
+/// This deliberately excludes the `read_*`/`glob`/`*_scan` family: those are how a
+/// catalog query legitimately reads Parquet and Iceberg over object storage.
+/// Reaching the *local* filesystem through them is refused by the session
+/// hardening instead.
+pub const MUTATING_OR_DYNAMIC_FUNCTIONS: &[&str] = &["nextval", "currval", "query", "query_table"];
```
konstantinoscs: You should add `json_execute_serialized_sql` here.

konstantinoscs (approved):
Add the function but otherwise lgtm! Thanks

bradhe:
## Review fixes landed in a23beed

All five findings are addressed. One correction on the SSRF finding, with data.

### P1 `json_execute_serialized_sql` — confirmed, fixed

Reproduced exactly as reported:

```
gate verdict for json_execute_serialized_sql  => Allowed
executed through json_execute_serialized_sql  => Ok("42")
gate verdict "SELECT * FROM enable_logging()" => Allowed
gate verdict "SELECT checkpoint()"            => Allowed
gate verdict "SELECT setseed(0.5)"            => Allowed
```

Self-inflicted, too: the `json` extension enabled to build the gate is what supplies the bypass. The deeper point, that a name denylist cannot be fail-closed, is the one I acted on. The gate now allowlists the **table-function position**, so an unknown function is refused rather than admitted, and asks DuckDB's own `has_side_effects` metadata about scalars so it keeps up with new versions without us maintaining a list. `has_side_effects` is NULL for every table function, which is exactly why that position is allowlisted instead. An explicit list covers what neither mechanism reaches: the dynamic-SQL executors and the effectful table functions.

This tightens the read path: `read_csv`, `read_parquet`, `glob` and that family are no longer permitted, since a catalog query reads base tables, which are not functions at all.

### P1 byte ceiling — confirmed, fixed

```
limit=1MiB, actually returned 50,000,000 bytes in one row, truncated=Some(Bytes)
```

Measured after admitting the row. Now measured first and the row is withheld, so an over-budget first row yields an empty truncated result. Blunt, but it is what the ceiling was for. Since that only bounds what the caller is handed, the read path now hardens with `Hardening::agent()`, which sets an engine memory limit for what a query spends before the first row exists.

### P1 SSRF — does not reproduce, but the test criticism was right

I could not reproduce an allowed `read_text('http://…')` making successful requests. I counted TCP connections to a listener I controlled rather than reading error strings. Control first, to prove the harness works:

```
harden=false  httpfs_loaded=true  net_reached=true   => OK rows=1
harden=true   httpfs_loaded=true  net_reached=false  => Permission Error: LocalFileSystem disabled
```

Then eleven vectors under hardening (`read_csv`, `read_text`, `read_blob`, `read_json_auto`, `read_csv_auto`, `read_parquet`, `parquet_scan`, `iceberg_scan`, replacement scan), including with the full production extension set of httpfs + iceberg + parquet loaded:

```
TOTAL internal connections under hardening: 0
```

If you have a working repro I would like it, since we may be on different hardening revisions or DuckDB builds.

**That said, both of your other points stand and I fixed them.** The block is incidental: it comes from `disabled_filesystems='LocalFileSystem'` breaking path resolution before HTTPFileSystem is consulted, not from any egress control, so it is not something to rely on. And the test genuinely passed for the wrong reason, since its helper never loaded `httpfs`. It now loads the same extensions `attach_statements` does, counts connections to a listener it owns, and includes a control run so it cannot pass vacuously. Egress remains open by design, because an attached Iceberg catalog is made of S3 reads; that is documented rather than papered over, and closing it needs a network boundary outside the process.

### P2 both confirmed, fixed

MinIO is pinned to `RELEASE.2025-09-07T16-13-09Z` instead of `latest`, and the test now skips only when Docker is genuinely unreachable, so a registry outage or broken image fails loudly instead of silently disabling a security test. The iceberg skip is narrowed the same way. Workspace `rust-version` is corrected to 1.88, which `rust-toolchain.toml` and the dependency tree have required for a while; the old 1.81 was a promise the workspace could not keep.

bradhe:
## Correction: the SSRF finding was right, and I was wrong

I said above it did not reproduce. That was a false negative from testing only on macOS. CI failed on Linux and proved it:

```
assertion `left == right` failed: a hardened query reached the network
  left: 6
 right: 3
```

Three hardened queries reached the probe listener on Ubuntu. The block I measured on macOS was incidental: disabling `LocalFileSystem` breaks HTTP path resolution before the HTTP filesystem is consulted, and Linux does not do that. So the data I posted was accurate and the conclusion drawn from it was not, because one platform is not the platform. Apologies for the pushback.

Your underlying point was the correct one throughout: the hardening is not an egress control, and treating it as one was the mistake.

### Fixed in e337d08

Two real holes, both closed at the gate, which is platform-independent:

**Replacement scans.** `SELECT * FROM 'http://…'` parses as an ordinary base table carrying the URL as its name, so the table-function allowlist never saw it. This was the live vector, and it is exactly the case you named. Base-table references that name a file or URL are now refused, with their own verdict so the message can say what is actually wrong.

**`iceberg_scan` accepted a location** and was on my allowlist, leaving another way to reach internal services. It is out. Every remaining entry (`range`, `generate_series`, `unnest`) is a pure generator taking no path and no URL, and the doc comment now states that adding anything which takes a location reopens this.

The test also asserted the wrong thing. It executed SQL directly, so it was testing the hardening in isolation, which this crate's own docs say does not close egress. It now mirrors production by running the gate first and executing only what the gate allows. A companion test, `hardening_alone_does_not_close_network_egress`, records the platform-dependent behaviour so the next person does not rediscover it in CI.

Verified on Linux: `45 passed, 0 failed, 0 ignored` in tower-duckdb, with the SSRF test doing real work rather than skipping, and the MinIO integration test running.

### Note for reviewers

Read mode is now noticeably tighter than earlier in this PR. No `read_csv`/`read_parquet`/`glob`, no `iceberg_scan`, and no `FROM 'some/path'`. A read-only catalog query reads the attached catalog's own tables; anything else needs `--write`. Deliberate, and the right default for an agent-facing path, but worth flagging as a behaviour change.

PR #336, opened 2026-07-30
refactor: show beta notices whenever stderr is a terminal

socksy:
In keeping with #322 this refactors the beta code to not use stdout/stderr stuff directly from non-`output.rs` related code

## Summary by CodeRabbit

* **Improvements**
  * Beta feature notices are now displayed consistently across storage and knowledge commands.
  * Notices appear only when appropriate and are shown no more than once, reducing repeated messages.
  * Beta notices continue to include relevant labels and documentation links when available.

### DIFF
--- a/crates/tower-cmd/src/output.rs
+++ b/crates/tower-cmd/src/output.rs
@@ -616,8 +616,23 @@ pub fn background_error(msg: &str) {
     write_to_stderr(&format!("{} {}\n", "Oh no!".red(), msg));
 }
 
+/// Writes a labelled notice to stderr once per user, ever. The once-per-user
+/// claim is only spent when stderr is a terminal, so a script, MCP capture, or
+/// CI run can't use it up on a notice nobody saw.
+pub(crate) fn notice_once(id: &str, label: &str, msg: &str) {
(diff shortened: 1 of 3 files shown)


crates/tower-cmd/src/output.rs
```diff
@@ -616,8 +616,23 @@ pub fn background_error(msg: &str) {
...
+/// CI run can't use it up on a notice nobody saw.
+pub(crate) fn notice_once(id: &str, label: &str, msg: &str) {
+    if !io::stderr().is_terminal() {
+        return;
+    }
+
+    match config::claim_notice(id) {
+        Ok(true) => notice_to_stderr(label, msg),
+        Ok(false) => {}
+        Err(err) => debug!("Failed to persist CLI notice {}: {}", id, err),
```
socksy, line 630: not realllly sure we need this, but wanted to keep the diff here clean as a move

konstantinoscs: This writes locally (similar to session persistence). Keep it

konstantinoscs (approved):

PR #346, opened 2026-08-13
ci: test minimum and latest Iceberg dependencies

konstantinoscs:
## What this does

Adds a small compatibility matrix for the Tables SDK dependencies:

- tests the exact minimum versions currently promised: PyArrow 23.0.1 and PyIceberg 0.11.1
- tests the latest stable versions selected by the package resolver
- runs on every PR and once daily, so new upstream releases are checked even when the CLI repository is quiet
- prints the versions installed in each job

This PR does not add upper bounds. The existing `>=` requirements remain unchanged, so users can install newer stable releases.

## Why

`pyproject.toml` states which versions are installable; this matrix verifies that the oldest promised versions and the newest available versions actually work. If an upstream release introduces a real incompatibility, CI will show it and we can add a temporary cap based on evidence.

The focused suite also keeps the existing managed catalog, S3 Tables, and BYO catalog behavior under test.

## Stack

PR 2 of 5. #345 is merged. Next: #347.

### DIFF
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -7,6 +7,9 @@ name: "[tower] Test python"
 
 on:
   pull_request:
+  schedule:
+    # Detect new stable PyArrow/PyIceberg releases even when no PR is open.
+    - cron: "17 6 * * *"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -14,6 +17,7 @@ concurrency:
 
 jobs:
   test:
+    if: github.event_name == 'pull_request'
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/test-python.yml
```diff
@@ -43,3 +43,66 @@ jobs:
...
+          assert Version("23.0.1") <= pyarrow < Version("26")
+          assert Version("0.11.1") <= pyiceberg < Version("0.12")
+
+          if os.environ["BOUNDS"] == "minimum":
+              assert pyarrow == Version("23.0.1")
+              assert pyiceberg == Version("0.11.1")
+          PY
+
+      - name: Run Iceberg compatibility tests
+        run: uv run --no-sync pytest tests/tower/test_tables.py tests/tower/test_storage.py
```
bradhe, line 98: What is this doing exactly? I don't understand why we need something in CI to check this? The rules are pretty well understood in `pyproject.toml`.

pyproject.toml
```diff
@@ -48,7 +48,12 @@ tower = "tower.cli:main"
 
 [project.optional-dependencies]
 ai = ["huggingface-hub>=0.34.3", "ollama>=0.5.3"]
-iceberg = ["polars>=1.39.3", "pyarrow>=23.0.1", "pyiceberg>=0.11.1", "boto3>=1.35.0"]
+iceberg = [
+  "polars>=1.39.3",
+  "pyarrow>=23.0.1,<26",
+  "pyiceberg>=0.11.1,<0.12",
```
bradhe: Feels like this is a compatibility trap to me--are we going to break someone?

konstantinoscs:
@bradhe take another look. I rethought about this. My initial idea was to bound pyiceberg and arrow to make explicit what version we support and what not as unbounded updates can theoretically break the CLI. However I think we can achieve a quite good level of support without bounding the dependencies by triggering a daily job that basically tests that we still support the two pairs of lowest/lowest and latest/latest pyiceberg and arrow. That gives us a quite early signal if there's ever a breaking change upstream

.github/workflows/test-python.yml
```diff
@@ -7,13 +7,17 @@ name: "[tower] Test python"
 
 on:
   pull_request:
+  schedule:
+    # Detect new stable PyArrow/PyIceberg releases even when no PR is open.
+    - cron: "17 6 * * *"
```
bradhe, line 12: Should this just be a separate workflow? Do we need to run the tests on this interval?

konstantinoscs: This just runs the tests once a day as well as on every PR.
To me it seems it belongs here. We should run the tests + dependencies compatibility at once? Seems like a reasonable "all things python" coupling.
But If you feel strongly about it, I can move it to another workflow I guess

bradhe (approved):
LGTM

PR #347, opened 2026-08-13
fix: common mechanism for retries in insert, upsert and delete

konstantinoscs:
## What this does

This PR makes table writes more resilient when two writers try to commit at the same time.

Insert, upsert, and delete now share one retry mechanism. After a genuine Iceberg commit conflict, the SDK waits for a small randomized delay, refreshes the table metadata, and tries again. The delay grows between attempts and is capped at 30 seconds.

## Safety

Only a known `CommitFailedException` is retried. Authentication, permission, network, validation, not-found, and unknown-commit-state errors still fail immediately, because replaying those operations could duplicate or hide a write.

This does not change catalog routing or credential vending, so S3 Tables and BYO behavior is preserved.

## Testing

Tests cover all three write operations, retry timing and ordering, exhausted retries, refresh failures, and errors that must not be retried.

### DIFF
--- a/src/tower/_tables.py
+++ b/src/tower/_tables.py
@@ -1,15 +1,16 @@
 from __future__ import annotations
 
+import math
 import os
+import random
+import time
 from dataclasses import dataclass
-from typing import List, Optional, TypeVar, Union
+from typing import Callable, List, Optional, TypeVar, Union
 
 from pyiceberg.exceptions import CommitFailedException, NoSuchTableError
 
 TTable = TypeVar("TTable", bound="Table")
-
-import random
-import time
+TRetryResult = TypeVar("TRetryResult")
 
 import polars as pl
 import pyarrow as pa
@@ -36,6 +37,8 @@ from .utils.tables import (
     make_table_name,
     namespace_or_default,
 )
+
+_MAX_COMMIT_RETRY_DELAY_SECONDS = 30.0
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


src/tower/_tables.py
```diff
@@ -249,8 +252,30 @@ def rows_affected(self) -> RowsAffectedInformation:
     def _validate_retry_args(max_retries: int, retry_delay_seconds: float) -> None:
         if max_retries < 0:
             raise ValueError("max_retries must be >= 0")
-        if retry_delay_seconds < 0:
-            raise ValueError("retry_delay_seconds must be >= 0")
+        if not math.isfinite(retry_delay_seconds) or retry_delay_seconds < 0:
+            raise ValueError("retry_delay_seconds must be finite and >= 0")
```
jo-sm, line 256: I would suggest to give an upper bound (maybe 300 seconds / 5 minutes?), rather than only checking if the delay is finite.

src/tower/_tables.py
```diff
@@ -249,8 +252,30 @@ def rows_affected(self) -> RowsAffectedInformation:
...
+
+    def _commit_with_retry(
+        self,
+        operation: Callable[[], TRetryResult],
+        max_retries: int,
+        retry_delay_seconds: float,
```
jo-sm: I would suggest to use an `int` here, rather than `float`, unless there's a strong need for sub second delays.

src/tower/_tables.py
```diff
@@ -249,8 +252,30 @@ def rows_affected(self) -> RowsAffectedInformation:
...
+    ) -> TRetryResult:
+        retry_ceiling = min(retry_delay_seconds, _MAX_COMMIT_RETRY_DELAY_SECONDS)
+
+        for attempt in range(max_retries + 1):
+            try:
+                return operation()
+            except CommitFailedException:
+                if attempt == max_retries:
+                    raise
```
jo-sm, line 273: I would raise a new exception, like `CommitRetryExhaustionException`.

src/tower/_tables.py
```diff
@@ -249,8 +252,30 @@ def rows_affected(self) -> RowsAffectedInformation:
...
+
+                delay = random.uniform(0.0, retry_ceiling)
```
jo-sm: Hmm. The way that this is used, compared to what the naming of the variable is, could be confusing (I was not sure why we have this jitter when we're using `retry_delay_seconds`). I'd suggest to rename `retry_delay_seconds` to `max_retry_delay` to make it clear it's the maximum, not the actual delay.

src/tower/_tables.py
konstantinoscs: So the default value is itself a float (0.5). Also the randomized backoff produces fractional values.
I think float is the correct thing here

src/tower/_tables.py
konstantinoscs: It's a bit confusing because the upper bound is anyway decided in `retry_ceiling = min(retry_delay_seconds, _MAX_COMMIT_RETRY_DELAY_SECONDS)` inside `_commit_with-retry_()`.
I'll add some clarification in the documentation for this

src/tower/_tables.py
konstantinoscs: This should remain `CommitFailedException` because it's actually the PyIceberg exception.
Preserving that exception:
1. retains the original failure details;
2. keeps existing except CommitFailedException handlers working;
3. avoids introducing Tower-specific behavior for S3 Tables and BYO callers.

A new exception would be a breaking change unless it subclassed `CommitFailedException`, and even then it adds little value.

src/tower/_tables.py
konstantinoscs: Actually, `max_retry_delay` is also inaccurate: the supplied value is only the maximum for the first retry; later ceilings grow until the separate 30-second cap we discussed above.

The biggest problem though is that It is also an existing public keyword across `insert`, `upsert`, and `delete`, so renaming it breaks callers on every catalog type.
I’d keep `retry_delay_seconds`, improve its documentation, and use initial_retry_ceiling_seconds internally

src/tower/_tables.py
jo-sm: If it's handled somewhere else, I would just be consistent with the above line then - `retry_delay_seconds < 0` is probably enough here.

socksy (approved):
ah nice

PR #348, opened 2026-08-13
fix: build delete filters with PyIceberg expressions

konstantinoscs:
## What this does

This PR replaces Tower's fragile PyArrow-filter string parser with native PyIceberg expressions.

Users can now write filters such as:

```python
table.delete(
    (table.column("age") >= 18)
    & (table.column("status") != "inactive")
)
```

SQL-like strings and native PyIceberg expressions remain supported. Old PyArrow expressions now raise a clear migration error before any write credentials are requested.

## Why

The previous implementation tried to understand a filter by parsing its rendered text. That could misread column names or values containing words such as `and`, `or`, or `not`. Building the expression structurally avoids those errors.

S3 Tables and BYO catalog selection and credentials are unchanged.

## Testing

Tests cover every comparison operator, logical composition and inversion, nested fields, unusual identifiers and values, and the PyArrow migration error.


bradhe (approved):
Love it! PyIceberg has come a long way from when we started this.

PR #349, opened 2026-08-13
fix: delegate Arrow schema conversion to PyIceberg

konstantinoscs:
## What this does

This PR passes a `pyarrow.Schema` directly to PyIceberg when creating a table and removes Tower's custom schema converter.

PyIceberg now handles field IDs, nested fields, documentation, nullability, and type validation using its standard public API.

## Why

Tower's converter duplicated PyIceberg behavior and could silently lose information, particularly timestamp and decimal precision. The native conversion is easier to maintain and rejects unsupported or lossy schemas clearly.

Tests also create tables through S3 Tables and BYO-style catalog configuration, confirming that those paths still use their existing PyIceberg catalog and never request Tower-managed credentials.

## Compatibility

Some schemas that were previously accepted after a lossy conversion now raise PyIceberg's normal validation errors. The reference documentation lists the supported timestamp, time, date, decimal, and timezone behavior.

## Testing

- 20 schema-focused tests pass.
- The complete stack passes 225 tests, with 10 skipped.
- Black and mypy pass.


bradhe (approved):
LGTM

PR #352, opened 2026-08-14
Add contributor licence agreement and enforcement workflow

bradhe:
Adds a Contributor License Agreement requirement for pull requests.

- `CLA.md`: individual CLA based on the Apache ICLA v2.2, with open items marked for legal review
- `.github/workflows/cla.yml`: CLA Assistant Lite workflow (pinned to v2.6.1 by commit SHA); signatures are stored in-repo under `signatures/version1/`
- `CONTRIBUTING.md`: short section explaining the signing flow

Note: the workflow uses `pull_request_target`, so enforcement only takes effect once this reaches `main`. Making the CLA status check required on `main` (branch protection) is a separate manual step.

### DIFF
--- a/.github/cla-allowlist.txt
+++ b/.github/cla-allowlist.txt
@@ -1,0 +1,25 @@
+# Accounts exempt from the CLA check (.github/workflows/cla.yml).
+# One GitHub username per line; lines starting with # and blank lines are
+# ignored. '*' wildcards are supported (e.g. bot accounts).
+#
+# The workflow reads this file from the default branch, so changes take
+# effect once they land there — a PR cannot allowlist its own author.
+#
+# Bots:
+dependabot[bot]
+renovate[bot]
+
+# Tower employees: exempt because they have already signed equivalent
+# agreements as part of employment; the CLA record for them lives in their
+# employment paperwork, not in signatures/. Keep in sync with org
+# membership; remove people when they leave (their later contributions need
+# a signature like anyone else's).
+bradhe
+datancoffee
+giray123
+jo-sm
+konstantinoscs
+MemoAlfa
+mesmith027
+sammuti
+socksy
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -29,29 +29,35 @@ permissions:
 jobs:
   cla:
     runs-on: ubuntu-latest
+    # The `github.event.issue.pull_request` guard is required: without it,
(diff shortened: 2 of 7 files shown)


jo-sm (commented):
A couple questions from my side:

1. If someone writes the phrase in a comment in PR A, and that PR gets closed, will then PR B require the phrase again? 
2. Is there a way to allowlist us who are in the organization? I don't see anything about it in the PR (maybe the action takes care of it?), but I think it would be a bit cumbersome if anyone who does something in the CLI who's working for Tower has to sign off first.

bradhe:
> If someone writes the phrase in a comment in PR A, and that PR gets closed, will then PR B require the phrase again?

Yeah until `signatures/version1/cla.json` gets updated, which I think we'll need to do operationally in the PR before we land it. There are more complicated solutions, but given the volume of third-party contributions we get, I don't think it's worth adopting for now 'till it becomes a problem.

Versioning the file also makes it trivially auditable, which I like.

> Is there a way to allowlist us who are in the organization?

I don't see a feature for this in the upstream. I'll attach a file to this PR with everyone in it already, and we'll just need to remember to do that as part of onboarding at Tower. Also, we'll get reminders, obviously, in the PRs that people open...

socksy (approved):

konstantinoscs (changes requested):
Just two things on my side:

1. `cla.json` targets main but at the same time main is protected. Shouldn't CLAs go through the develop -> main route like everything else?
2. We should require the CLA on develop, not main. That's where contributions will be

Also, funny enough, I think that the way we check for the "agreement" can be tricked with a phrase like “I hereby sign … but I do not agree” which will be accepted

bradhe:
Ah it looks like the action will directly update `cla.json` via the GitHub Contents API, so to @jo-sm 's question, it won't be necessary to update it manually theoretically (we should try this out). But anyway, the question is where should the content be stored. `develop` seems like the right place as it's the default branch. We'll need to slightly loosen our branch protection rules for this, I'm working on it.

With regard to where the CLA is required, it's basically everywhere: We don't set a target branch.

Claude's developed a solution to your poisoned comment topic, too. Good find.

konstantinoscs (approved):
Lgtm!


Also, two comments from Sol, not very important to me to block on but I leave them here:

1. Add set -euo pipefail before the allowlist pipeline so API/decoding failures cannot silently produce an empty list.

2. Confirm the company rename: the PR changes the license to GmbH, while pyproject.toml and flake.nix still say Inc.

PR #353, opened 2026-08-14
Remove third-party contributions and plug the gaps

bradhe:
This pull requests reverts third-party commits that aren't covered by our CLA and patches the holes left behind in the CLI. We have a separate process that we've introduced for tracking CLA signing going forward, we just can't reasonably do the CLA signing retroactively given the age of the commits.

## Why?

We ship `tower/tower-cli` under an MIT license and we have to show that every line we ship is actually ours to license. One contributor's commits carried an employment question we couldn't resolve in time, so leaving the code in place would have meant shipping something a third party might have a claim on, which is a risk that compounds with every release and every downstream user.

Reverting the code and rebuilding it from a functional spec, with the record to prove the separation, remediates the issue.


socksy (approved):
kinda hard to review in one go, but it all looks ok... I haven't tried it out though, putting some faith into those newly rewritten BDD tests (might be a good idea to run them against the real backend too?)

PR #359, opened 2026-08-19
fix: harden private Storage control-plane authentication

konstantinoscs:
## Summary

- Replace Storage's use of the shared legacy client with a private resolver for Tower catalog describe and credential-vend requests.
- Snapshot the established environment-backed authentication policy once per operation: `TOWER_JWT` first, then `TOWER_API_KEY`.
- Verify HTTPS certificates by default, keep explicit `http://` Tower URLs usable for local development, and use bounded request timeouts.
- Raise typed errors for missing, placeholder, and connection failures.
- Create and close a fresh generated HTTP client for every control-plane request, with no public client or close lifecycle.

## Why

This gives the Storage path one consistent authentication and request policy without changing the shared SDK client, the CLI, or CLI sessions. It is the private foundation for one-shot catalog loading and for a later managed `TableReference` to retain only resolution/cache state—not network resources.

The existing process-wide credential cache remains in this PR and is replaced by the next PR in the stack.

## Testing

- 61 focused Storage and Tables tests
- Black, Ruff, mypy, and diff checks

### DIFF
--- a/src/tower/_storage.py
+++ b/src/tower/_storage.py
@@ -77,19 +77,16 @@ def _auth_from_context(context: TowerContext) -> tuple[str, str, str]:
     return token, auth_header_name, prefix
 
 
-def _build_tower_control_plane_client(
+def _new_tower_control_plane_client(
     *,
-    context: TowerContext,
-    tower_url: str,
+    base_url: str,
+    token: str,
+    auth_header_name: str,
+    prefix: str,
     timeout: float,
-    verify_tls: bool,
-) -> tuple[str, str, AuthenticatedClient]:
-    token, auth_header_name, prefix = _auth_from_context(context)
-    base_url = _api_base_url(tower_url)
-    # hash whatever auth token was provided so we can cache catalogs tokens per account.
-    auth_hash = hashlib.sha256(token.encode("utf-8")).hexdigest()
-    client = AuthenticatedClient(
-        verify_ssl=verify_tls,
+) -> AuthenticatedClient:
+    return AuthenticatedClient(
+        verify_ssl=True,
         base_url=base_url,
         token=token,
         auth_header_name=auth_header_name,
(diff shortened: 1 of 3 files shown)


src/tower/_storage.py
```diff
@@ -33,13 +39,150 @@
...
+        try:
+            return vend_catalog_credentials_api.sync(
+                name=name,
+                client=self._tower_client,
+                environment=self.environment,
+                body=body,
+            )
+        except httpx.RequestError as error:
+            raise StorageConnectionError(
+                f"Could not connect to Tower at {self._base_url!r}."
```
jo-sm: I would say the `!r` is unnecessary here, unless you're following the convention from another file?

src/tower/_storage.py
```diff
@@ -33,13 +39,150 @@
...
+            ) from error
+
+
+def _api_base_url(tower_url: str) -> str:
+    try:
+        url = httpx.URL(tower_url)
+    except (TypeError, httpx.InvalidURL) as error:
+        raise ValueError(f"Invalid Tower URL: {tower_url!r}") from error
```
jo-sm: Same as the above, I'd just use omit the `!r`.

src/tower/_storage.py
```diff
@@ -33,13 +39,150 @@
...
+    if not url.is_absolute_url or url.scheme not in ("http", "https"):
+        raise ValueError(f"Invalid Tower URL: {tower_url!r}")
+    return str(url.copy_with(path="/v1", query=None, fragment=None)).rstrip("/")
```
jo-sm: Is there a way to build the URL so that you don't need to remove the trailing slash(es)?

src/tower/_storage.py
```diff
@@ -76,18 +219,18 @@ def get_tower_catalog_credentials(
...
+    cache_key = _cache_key(storage_client, name, mode)
 
     now = datetime.now(timezone.utc)
     _prune_credential_cache(now)
     cached = _credential_cache.get(cache_key)
     if cached is not None and cached.is_usable(now):
         return cached.credentials
 
-    credentials = _vend_with_default_catalog_fallback(ctx, name, environment, mode)
+    with storage_client._tower_client:
```
jo-sm: Does this `with` statement do anything given that this function is synchronous and the enter/exit functions of `AuthenticatedClient` are asynchronous?

src/tower/_storage.py
```diff
@@ -127,26 +273,19 @@ def _vend_with_default_catalog_fallback(
...
 ) -> str | None:
-    if not (ctx.api_key or ctx.jwt):
+    if ctx.jwt is None and ctx.api_key is None:
         return None
 
-    cache_key = (ctx.tower_url, name, environment)
+    base_url, auth_hash, tower_client = _build_tower_control_plane_client(
+        context=ctx,
+        tower_url=ctx.tower_url,
+        timeout=CATALOG_TYPE_DESCRIBE_TIMEOUT_SECONDS,
```
jo-sm: Maybe a more appropriate name would be `CONTROL_PLANE_CLIENT_DEFAULT_TIMEOUT`?

src/tower/_storage.py
konstantinoscs: Done!

src/tower/_storage.py
konstantinoscs: Done!

src/tower/_storage.py
konstantinoscs: It was useless anyway so I removed it.

src/tower/_storage.py
konstantinoscs: `AuthenticatedClient` has synchronous __enter__/__exit__ methods that close its `httpx.Client`; async methods are separate so this is needed

src/tower/_storage.py
konstantinoscs: This is a specific timeout for the Catalog describe operation. Normal storage requests have a separate 30s timeout

src/tower/_storage.py
jo-sm: Ah sorry I missed them, they were in the middle of the class definition.

jo-sm (approved):

PR #361, opened 2026-08-19
feat: cache resolved catalog access per private resolver

konstantinoscs:
## Summary

- Preserve the catalog owner returned with vended credentials, together with target environment, mode, endpoint details, token, and expiry.
- Instead of a process-wide credential cache, the resolver now owns a credential cache and knows when those expire.
- Share one vend request between concurrent identical resolutions on that same resolver.
- Validate returned mode and environment before caching.
- Keep one-shot credential loads isolated: each top-level call creates and discards its own resolver, so independent calls never share cache state.

## Why

This establishes the private resolved-access boundary needed by the remaining Phase 1 work. A future Tower-managed `TableReference` can retain one resolver and reuse only its own credentials, while the public one-shot catalog-loading UX stays stateless and exposes no client or close lifecycle.

Public exports and Tables integration remain out of scope for this PR. Owner-pinned read-to-write escalation is enforced when `TableReference` adopts this primitive later in the stack.

Visualization helper for `_resolve_catalog_access()`:
```
resolve(name, mode)

        │
        ▼
 normalize mode
        │
        ▼
check credential cache
        │
       valid?
      /      \
    yes       no
     │         │
   return      ▼
           Is somebody already
           fetching these creds?
              /       \
            yes        no
             │          │
             │       create Future
             │          │
         wait for       ▼
         Future      call Tower
                         │
                         ▼
                  validate response
                         │
                         ▼
                  create access object
                         │
                         ▼
                  cache if usable
                         │
                         ▼
                  complete Future
                         │
                         ▼
                       return
```

## Testing

- 66 focused Storage and Tables tests
- 240 full-suite tests passed, 10 skipped
- Black, Ruff, mypy, and diff checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved storage access caching to prevent duplicate requests during concurrent operations.
  * Automatically refreshes credentials that are close to expiring.
  * Validates returned environment and access mode information to detect inconsistent responses.
  * Retries access resolution after failed requests instead of retaining invalid results.

* **Tests**
  * Added coverage for caching, concurrent requests, credential refreshes, retries, and invalid responses.

### DIFF
--- a/src/tower/_storage.py
+++ b/src/tower/_storage.py
@@ -3,8 +3,10 @@ from __future__ import annotations
 import hashlib
 import logging
 import time
-from dataclasses import dataclass
+from concurrent.futures import Future
+from dataclasses import dataclass, field
 from datetime import datetime, timedelta, timezone
+from threading import Lock
 from typing import TYPE_CHECKING
 
 import httpx
@@ -15,6 +17,7 @@ if TYPE_CHECKING:
 from ._context import TowerContext
 from .exceptions import (
     StorageConnectionError,
+    StorageError,
     StorageInvalidCredentialError,
     StorageMissingAuthenticationError,
 )
@@ -51,6 +54,12 @@ INVALID_CREDENTIAL_SENTINEL = "<redacted>"
 logger = logging.getLogger("tower.storage")
 
 
+@dataclass(frozen=True, slots=True)
+class _AccessCacheKey:
+    name: str
+    mode: str
+
+
 def _auth_from_context(context: TowerContext) -> tuple[str, str, str]:
     if context.jwt is not None:
         token = context.jwt
@@ -69,7 +78,7 @@ def _auth_from_context(context: TowerContext) -> tuple[str, str, str]:
 
     if token.strip() == INVALID_CREDENTIAL_SENTINEL:
         raise StorageInvalidCredentialError(
-            f"{source} contains the {INVALID_CREDENTIAL_SENTINEL!r} placeholder, "
+            f"{source} contains the {INVALID_CREDENTIAL_SENTINEL} placeholder, "
             "not a usable Tower credential."
         )
     if not token.strip():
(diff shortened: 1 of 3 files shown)


src/tower/_storage.py
```diff
@@ -142,6 +174,79 @@ def _new_client(self) -> AuthenticatedClient:
...
+        self,
+        name: str,
+        mode: str,
+    ) -> _ResolvedCatalogAccess:
+        mode = _normalize_mode(mode)
+        target_environment = self._target_environment
+        # Host, authentication, and target are fixed for this resolver.
+        cache_key = (name, mode)
+
+        with self._access_lock:
```
jo-sm, line 193: Are you adding locking mechanisms to allow this to be used in a threaded context?

src/tower/_storage.py
```diff
@@ -142,6 +174,79 @@ def _new_client(self) -> AuthenticatedClient:
...
+        try:
+            response = _vend_with_default_catalog_fallback(
+                self,
+                name,
+                mode,
+            )
+            credentials = response.credentials
+            if credentials.mode != mode:
+                raise StorageError(
+                    f"Tower returned {credentials.mode!r} credentials after "
```
jo-sm: I'd suggest to remove the `!r` from here and other places.

src/tower/_storage.py
```diff
@@ -104,6 +110,34 @@ def _auth_hash(token: str, auth_header_name: str, prefix: str) -> str:
...
+    catalog_name: str
+    catalog_uri: str
+    warehouse: str
+    mode: str
+    oauth_token: str = field(repr=False)
+    expires_at: datetime
+
+    @property
+    def is_inherited(self) -> bool:
+        return self.catalog_environment != self.target_environment
```
jo-sm: I would suggest to just keep this as a function so that the interface is consistent with `is_usable`. I originally thought `is_usable` should be a property too until I saw it takes an argument, and since they have the same name I think keeping them consistent is preferable.

src/tower/_storage.py
konstantinoscs: Ah, this again heh. Will do!

src/tower/_storage.py
konstantinoscs: Yes. In case there are multiple credential requests involved, to avoid spamming our API (and our API spamming polaris to re-mint tokens).
I've added a helper diagram for how `resolve_catalog_access` is supposed to work

src/tower/_storage.py
konstantinoscs: Hmm, ok I don't feel strongly about it

konstantinoscs: Done

src/tower/_storage.py
konstantinoscs: Done

src/tower/_storage.py
jo-sm: Just to make sure, this is supposed to handle the case where a single script tries to do N operations, not separate scripts doing the same operations, right?

konstantinoscs: Yep haha exactly. Someone spawns a thread pool of workers doing table accesses

jo-sm (approved):

PR #365, opened 2026-08-20
feat: let the server decide the maximum bundle size

bradhe:
The CLI currently rejects every package over 50 MiB before uploading, even when the server allows larger bundles. This change removes that client-side cap and leaves enforcement and rejection details to the server.

Packages above 500 MiB produce a non-fatal warning in human output. The warning uses adaptive binary units (MiB/GiB) and is suppressed in JSON mode through `Out::note`, preserving machine-readable output. The advisory threshold is private to the deploy module, and the unused public `MAX_PACKAGE_SIZE` constant and re-export are removed.

Uploads remain streamed and now send the known package size in `Content-Length`, allowing servers to reject oversized requests from their headers. Existing server error rendering is unchanged. Older installed CLI versions retain the 50 MiB cap until upgraded.

Validation: the local HTTP regression test verifies the advertised length, streamed body, gzip header, and absence of chunked transfer encoding. It passes with `DEVELOPER_DIR=/Library/Developer/CommandLineTools cargo test -p tower-cmd upload_advertises_exact_streamed_length --lib`. Formatting of the changed file and `git diff --check` pass. Workspace-wide formatting reports pre-existing differences in unrelated files. Snyk Code scanning was attempted but is unavailable because it is disabled for the configured organization.

### DIFF
--- a/crates/tower-cmd/src/util/deploy.rs
+++ b/crates/tower-cmd/src/util/deploy.rs
@@ -16,8 +16,10 @@ use tower_api::models::DeployAppResponse;
 
 /// Advisory only: the server enforces the actual bundle-size limit and reports
 /// it in its error response when a bundle is too large.
-pub const LARGE_PACKAGE_WARNING_THRESHOLD: u64 = 500 * 1024 * 1024;
+const LARGE_PACKAGE_WARNING_THRESHOLD: u64 = 500 * 1024 * 1024;
 
+/// Streams a package to the server, reporting progress and leaving size enforcement
+/// to the server. Large-package warnings are suppressed in JSON output mode.
 pub async fn upload_file_with_progress(
     out: &output::Out,
     api_config: &Configuration,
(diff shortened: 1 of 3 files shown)


crates/tower-cmd/src/util/deploy.rs
```diff
@@ -36,13 +40,11 @@ pub async fn upload_file_with_progress(
...
-    // Check if bundle size exceeds the maximum allowed size
-    if file_size > tower_package::MAX_PACKAGE_SIZE {
+    if file_size > LARGE_PACKAGE_WARNING_THRESHOLD {
         let size_mb = file_size as f64 / (1024.0 * 1024.0);
-        let max_mb = tower_package::MAX_PACKAGE_SIZE as f64 / (1024.0 * 1024.0);
-        out.die(&format!(
-            "Your App is too big! ({:.2} MB) exceeds maximum allowed size ({:.0} MB). Please consider reducing app size by removing unnecessary files or import_paths in the Towerfile.",
-            size_mb, max_mb
+        out.write(&format!(
+            "Warning: Your app package is large ({:.2} MB). The server may reject it depending on its configured maximum bundle size. You can reduce app size by removing unnecessary files or import_paths in the Towerfile.\n",
```
jo-sm: How do you feel about making this a bit nicer to read, like `ls -lh` or `du -sh`, so that users will see something like `120 MB` or `1.2GB`?

crates/tower-cmd/src/util/deploy.rs
```diff
@@ -36,13 +40,11 @@ pub async fn upload_file_with_progress(
...
 
-    // Check if bundle size exceeds the maximum allowed size
-    if file_size > tower_package::MAX_PACKAGE_SIZE {
+    if file_size > LARGE_PACKAGE_WARNING_THRESHOLD {
         let size_mb = file_size as f64 / (1024.0 * 1024.0);
-        let max_mb = tower_package::MAX_PACKAGE_SIZE as f64 / (1024.0 * 1024.0);
-        out.die(&format!(
-            "Your App is too big! ({:.2} MB) exceeds maximum allowed size ({:.0} MB). Please consider reducing app size by removing unnecessary files or import_paths in the Towerfile.",
-            size_mb, max_mb
+        out.write(&format!(
```
konstantinoscs: I think this will break `--json`. I think `out.note` or even including this structurally as json will work

konstantinoscs (commented):
We already know the size of the bundle right? We could utilize sth like `Content-Length: file_size` to make early rejections

socksy (approved):

PR #371, opened 2026-09-22
feat: add run_size to the Towerfile

migueltheduck:
## Why

The API (v0.12.5) accepts `[app] run_size` and reads it from the Towerfile stored on each app version — but the CLI didn't model the field. Since `App` isn't `deny_unknown_fields`, a `run_size` line parsed fine and was dropped from the struct. Every MCP write goes read → mutate → `Towerfile::save()`, so **any `tower_file_update` / `tower_file_add_parameter` call silently deleted the user's `run_size`**, downgrading the app to the default size on the next deploy.

## What

- `App.run_size: Option<String>`, mirroring `description` (`skip_serializing_if`), so an unset value is never written into an existing Towerfile.
- Validation against the four known sizes (`basic.xsmall` / `small` / `medium` / `large`), so a typo fails at `tower deploy` with the valid list instead of server-side.
- The field checks move out of `from_toml()` into `Towerfile::validate()`, which `save()` now also calls. This fixes a second, latent bug: `save()` never re-validated, so the MCP tool would have written an invalid size to disk and reported success.
- `tower_file_update` gains `run_size`; the skill header documents it (SKILL.md regenerated, 2-line diff).


bradhe (approved):
Looks good to me

PR #372, opened 2026-09-22
fix: make `tower package` exit non-zero on failure

migueltheduck:
## Why

Every error branch in `do_package` reported the problem and then fell through to a zero exit. Reproduced for a missing Towerfile, a missing `script`, a script not in the bundle, and an unwritable output path — all returned 0. `tower deploy` already gets this right; `package` didn't. Anyone using the (hidden) `tower package` as a build step couldn't tell success from failure.

## What

`std::process::exit(1)` after each error report, matching `deploy.rs`. Three-line change to [package.rs](crates/tower-cmd/src/package.rs).

## Verification

- Manual: all four failure modes above now exit 1; a valid app still exits 0.
- New `tests/integration/features/cli_package.feature` (3 scenarios) plus generic `the command should exit with a non-zero status` / `with status 0` steps for reuse. Both non-zero scenarios fail against the pre-fix binary.

Found while working on #371; independent of it.


migueltheduck:
Disregard

PR #373, opened 2026-09-22
feat: preserve Towerfile keys the CLI does not model

migueltheduck:
> Stacked on #371 (base branch `feat/towerfile-run-size`). GitHub retargets this to `develop` once that merges.

## Why

Every MCP write goes read → mutate → `Towerfile::save()`, and `save()` re-serializes the struct — so any key the struct doesn't model is silently deleted from the user's file. `run_size` was one instance (#371). The `[build]` table in `tests/integration/templates/Towerfile.j2` is another. Any field a newer server adds before the CLI catches up would be a third. The server reads the raw Towerfile, so a client that destroys what it doesn't understand is actively harmful.

## What

`App` and `Towerfile` gain `extra: toml::Table` holding unknown keys / sections, carried through read → modify → save unchanged.

**Why not just `#[serde(flatten)]`?** It works for serializing, and that's how it's used here. For deserializing it does not: serde buffers flattened leftovers through its generic `Content` type, which turns TOML datetimes into `{ "$__toml_private_datetime" = "…" }` — and that then gets written back into the file. Worse, toml's own visitor recognises that marker on re-parse, so the corruption round-trips invisibly. `Deserialize` is therefore implemented by hand: parse as `toml::Table`, pop known keys with `#[serde(default)]` semantics, keep the rest intact. Wrong-typed known keys still error, naming the key (`app.name`).

**Why not `deny_unknown_fields`?** It would reject the integration template outright and make every older CLI fail on every newer Towerfile — the opposite of what a client should do with a file the server reads directly.

## Verification

- Unit: unknown `[app]` keys and top-level sections survive edit + save; a plain Towerfile serializes byte-identically (no `extra` leaks); datetimes are asserted **on the written text**, not the re-parsed value, for the reason above; wrong types and a missing `[app]` still fail clearly.
- behave (real MCP server): editing `description` leaves `[build] python = "3.11"` in place. Fails on `develop`.
- `cargo check --workspace --all-features`; tower-cmd's 121 tests unchanged.

## Not in scope

Comments and formatting are still rewritten on every save. Doing that properly means editing in place with `toml_edit` — a separate change.


migueltheduck:
Disregard

PR #374, opened 2026-09-22
feat: add --run-size to `tower run` and `tower_run_remote`

migueltheduck:
> Stacked on #371 (base branch `feat/towerfile-run-size`); sibling of #373. GitHub retargets this to `develop` once #371 merges. **Review commit by commit** — the first is a generated-code bump, the second is the change.

## Why

#371 gives an app a default size in its Towerfile. The API also accepts a per-run override on `RunAppParams.run_size`, which the CLI had no way to send — the vendored client predated the field.

## Commit 1 — `chore: regenerate tower-api client from spec v0.12.6`

248 files, but almost entirely the per-file header comment bumping `v0.11.24 → v0.12.6`. The substantive changes:

- `RunAppParams.run_size: Option<RunSize>` — new enum of the four managed sizes.
- `AppVersion.run_size: String` — **required**; the size the version's Towerfile declared. The mock API server now returns it, otherwise the regenerated client can't deserialize a deploy response.
- `describe_default_catalog` (`GET /storage/catalogs/default`) was removed from the spec. Nothing in tower-cmd used it.

Generated with openapi-generator-cli 7.19.0 using the repo's own config and templates, run through the official Docker image (the script wants a local JDK + Maven). Diff is what the script would produce. 32 model files were already orphaned from `models/mod.rs` before this; left alone.

## Commit 2 — the flag

- `tower run --run-size <SIZE>`, validated by clap against `tower_package::RUN_SIZES`, so a typo gets the list of valid names. `conflicts_with("local")` — a compute size means nothing for a local run.
- `tower_run_remote` gains optional `run_size`; unknown values are rejected before the API call.
- `parse_run_size` maps the string onto the generated `RunSize` (which has no `Deserialize`). A unit test round-trips every `RUN_SIZES` entry through the enum's serde name so the Towerfile list and the API enum can't drift.
- SKILL.md regenerated (one line).

## Verification

- `cargo test --all-features` green across the workspace (125 in tower-cmd, 4 new).
- behave: 6 new scenarios in `cli_run_size.feature` / `mcp_run_size.feature`, tagged `@serial`. They assert on the mock's new `/test/run-log` — i.e. on what was actually sent — for an explicit size, no size, an unknown size, and `--local` + `--run-size`. Full suite: 50 parallel + 13 serial, 0 failed.

## Not in this PR

- `tower apps show` could display `AppVersion.run_size` now that it's in the model. Small, separate.
- The Python SDK's `run_app()` and its generated client (`src/tower/tower_api_client`) don't know about `run_size` yet — that's `scripts/generate-python-api-client.sh` plus a kwarg, its own change.


migueltheduck:
Disregard

